PR #1, opened 2024-05-13
Add project skeleton

shadeMe:


julian-risch (changes requested):
Let's copy also the [gitignore from Haystack except for the section with Haystack files](https://github.com/deepset-ai/haystack/blob/main/.gitignore) and remove the .DS_Store file

test/conftest.py
```diff
@@ -0,0 +1,95 @@
...
+    return tokenizer
+
+
+@pytest.fixture()
+def test_files_path():
+    return Path(__file__).parent / "test_files"
+
+
+@pytest.fixture
+def mock_chat_completion():
```
julian-risch, line 37: Is this part of the PR for a specific reason?

shadeMe: Ah, no. Removed.

julian-risch (approved):

PR #4, opened 2024-05-16
feat: Implement `PipelinePair` primitive and related utility functions

shadeMe:
This is the first of a series of PRs that revolves around implementing an evaluation harness (c.f https://github.com/deepset-ai/haystack/issues/7526).

`PipelinePair` is a utility class that allows for two pipelines to be used in tandem. The motivation behind this abstraction is  to reduce the amount of glue code it takes to connect a, say, RAG pipeline with an evaluation pipeline. It aims to be generic enough work with arbitrary pipelines while making special accommodations to ensure that pipelines of different "shapes" can be connected without much extra work from the user. "Shape" here refers to the modality of pipeline inputs and outputs, e.g: a typical RAG pipeline takes a query and returns an answer whereas a typical RAG evaluation pipeline will take batches of queries and answers. Connecting two such pipelines is handled by automatically aggregating and deaggregating inputs and outputs at the interface between the pipelines.

Blocked by #2.

### DIFF
--- a/haystack_experimental/evaluation/pipeline_pair.py
+++ b/haystack_experimental/evaluation/pipeline_pair.py
@@ -1,0 +1,197 @@
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple
+
+from haystack import Pipeline
+
+
+@dataclass(frozen=True)
+class PipelinePair:
+    """
+    A pair of pipelines that are linked together and
+    executed sequentially.
+
+    :param first:
+        The first pipeline in the sequence.
+    :param second:
+        The second pipeline in the sequence.
+    :param outputs_to_inputs:
+        A mapping of the outputs of the first pipeline to the
+        inputs of the second pipeline in the following format:
+        `"name_of_component.name_of_output": "name_of_component.name_of_input`.
+        A single output can be mapped to multiple inputs.
+    :param map_first_outputs:
+        A function that post-processes the outputs of the first
+        pipeline, which it receives as its (only) argument.
+    :param included_first_outputs:
+        Names of components in the first pipeline whose outputs
+        should be included in the final outputs.
+    :param included_second_outputs:
+        Names of components in the second pipeline whose outputs
+        should be included in the final outputs.
+    """
+
+    first: Pipeline
+    second: Pipeline
+    outputs_to_inputs: Dict[str, List[str]]
+    map_first_outputs: Optional[Callable] = None
+    included_first_outputs: Optional[Set[str]] = None
+    included_second_outputs: Optional[Set[str]] = None
+
+    def __post_init__(self):
+        first_outputs = self.first.outputs(include_components_with_connected_outputs=True)
+        second_inputs = self.second.inputs(include_components_with_connected_inputs=True)
+        seen_second_inputs = set()
+
+        # Validate the mapping of outputs from the first pipeline
+        # to the inputs of the second pipeline.
+        for first_out, second_ins in self.outputs_to_inputs.items():
+            first_comp_name, first_out_name = self._split_input_output_path(first_out)
+            if first_comp_name not in first_outputs:
+                raise ValueError(f"Output component '{first_comp_name}' not found in first pipeline.")
+            elif first_out_name not in first_outputs[first_comp_name]:
+                raise ValueError(
+                    f"Component '{first_comp_name}' in first pipeline does not have expected output '{first_out_name}'."
+                )
+
+            for second_in in second_ins:
+                if second_in in seen_second_inputs:
+                    raise ValueError(
+                        f"Input '{second_in}' in second pipeline is connected to multiple first pipeline outputs."
+                    )
+
+                second_comp_name, second_input_name = self._split_input_output_path(second_in)
+                if second_comp_name not in second_inputs:
+                    raise ValueError(f"Input component '{second_comp_name}' not found in second pipeline.")
+                elif second_input_name not in second_inputs[second_comp_name]:
+                    raise ValueError(
+                        f"Component '{second_comp_name}' in second pipeline does not have expected input '{second_input_name}'."
+                    )
+                seen_second_inputs.add(second_in)
+
+    def _validate_second_inputs(self, inputs: Dict[str, Dict[str, Any]]):
+        # Check if the connected input is also provided explicitly.
+        second_connected_inputs = [
+            self._split_input_output_path(p_h) for p in self.outputs_to_inputs.values() for p_h in p
+        ]
+        for component_name, input_name in second_connected_inputs:
+            provided_input = inputs.get(component_name)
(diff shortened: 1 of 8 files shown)


haystack_experimental/evaluation/pipeline_pair.py
```diff
@@ -0,0 +1,197 @@
...
+    def _split_input_output_path(self, path: str) -> Tuple[str, str]:
+        # Split the input/output path into component name and input/output name.
+        pos = path.find(".")
+        if pos == -1:
+            raise ValueError(
+                f"Invalid pipeline i/o path specifier '{path}' - Must be in the following format: <component_name>.<input/output_name>"
+            )
+        return path[:pos], path[pos + 1 :]
+
+    def _prepare_reqd_outputs_for_first_pipeline(self) -> Set[str]:
```
davidsbatista, line 94: typo

```suggestion
    def _prepare_read_outputs_for_first_pipeline(self) -> Set[str]:
```

shadeMe: Ah, not really - `reqd` == `required`

haystack_experimental/evaluation/pipeline_pair.py
```diff
@@ -0,0 +1,197 @@
...
+            first_component, first_output = self._split_input_output_path(first_output)
+
+            # Each output from the first pipeline can be mapped to multiple inputs in the second pipeline.
+            for second_input in second_input_candidates:
+                second_component, second_input_socket = self._split_input_output_path(second_input)
+
+                second_component_inputs = second_inputs.get(second_component)
+                if second_component_inputs is not None:
+                    # Pre-condition should've been validated earlier.
+                    assert second_input_socket not in second_component_inputs
```
davidsbatista, line 116: I would remove this `assert`— the check was already done. If it's here to inform whoever reads the code, it can be replaced by a comment.

In the case it's here to guarantee that we are not overwriting and already added `second_input_socket` I suggest a try/except informing the reason for the exception.

haystack_experimental/evaluation/pipeline_pair.py
```diff
@@ -0,0 +1,197 @@
...
+        for component_name, input_name in second_connected_inputs:
+            provided_input = inputs.get(component_name)
+            if provided_input is None:
+                continue
+            elif input_name in provided_input:
+                raise ValueError(
+                    f"Second pipeline input '{component_name}.{input_name}' cannot be provided both explicitly and by the first pipeline."
+                )
+
+    def _split_input_output_path(self, path: str) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
shadeMe: I've made it a static method for now. If we see a need for this function elsewhere, we can move it to `util.py`.

haystack_experimental/evaluation/pipeline_pair.py
```diff
@@ -0,0 +1,198 @@
...
+        # Split the input/output path into component name and input/output name.
+        pos = path.find(".")
+        if pos == -1:
+            raise ValueError(
+                f"Invalid pipeline i/o path specifier '{path}' - Must be in the following format: <component_name>.<input/output_name>"
+            )
+        return path[:pos], path[pos + 1 :]
+
+    def _prepare_reqd_outputs_for_first_pipeline(self) -> Set[str]:
+        # To ensure that we have all the outputs from the first
```
davidsbatista, line 96: I was a bit confused, a simple suggestion to better understand what the output here is:

```
get all the components from the first pipeline whose outputs:
 - are required by the second pipeline as input
 - whose outputs should be included in the final outputs
```

davidsbatista (approved):

haystack_experimental/evaluation/pipeline_pair.py
shadeMe: Let's keep the assertion as it's meant to enforce the expected pre-condition. Neither swapping that with a comment nor switching it with an exception we'd either lose the above or change the semantics of the function.

davidsbatista:
This was a rather a big one, I tried to cover as much detail as possible, the logic looks OK to me, but I might have missed some edge cases - maybe checking what the test coverage shows to be sure we are testing every relevant piece of code.

haystack_experimental/evaluation/pipeline_pair.py
shadeMe: I've expanded on the existing comment.

haystack_experimental/evaluation/pipeline_pair.py
```diff
@@ -0,0 +1,201 @@
...
+            first_inputs, include_outputs_from=self._prepare_reqd_outputs_for_first_pipeline()
+        )
+        if self.map_first_outputs is not None:
+            first_outputs = self.map_first_outputs(first_outputs)
+        second_inputs = self._map_first_second_pipeline_io(first_outputs, second_inputs)
+        second_outputs = self.second.run(second_inputs, include_outputs_from=self.included_second_outputs)
+
+        return {"first": first_outputs, "second": second_outputs}
+
+    def run_first_as_batch(
```
julian-risch, line 162: Let's get input from devrel regarding "batch" in the name and how to best explain this concept to users. We run the first pipeline here in a loop, one query at a time. The docstring is good and makes that clear. "batch execution", and "run_first_as_batch" however give the impression that we're processing batches and some users might wrongly expect performance gains from that.

haystack_experimental/evaluation/pipeline_pair.py
```diff
@@ -0,0 +1,201 @@
...
+        A single output can be mapped to multiple inputs.
+    :param map_first_outputs:
+        A function that post-processes the outputs of the first
+        pipeline, which it receives as its (only) argument.
+    :param included_first_outputs:
+        Names of components in the first pipeline whose outputs
+        should be included in the final outputs.
+    :param included_second_outputs:
+        Names of components in the second pipeline whose outputs
+        should be included in the final outputs.
```
julian-risch, line 30: Let's add a usage example here. Similar to what test_pipeline_pair_run_first_as_batch does but with real pipelines.

julian-risch (commented):
I agree this idea will simplify running a pipeline with the outputs of another pipeline as inputs. It looks quite flexible. 👍 Let's try it out with some examples and get feedback from users.

haystack_experimental/evaluation/pipeline_pair.py
shadeMe: FYI, this is not going to be part of the public API - It's currently meant to be used as an internal helper class for evaluation. Promoting it to a first class citizen is a discussion for the future.

haystack_experimental/evaluation/pipeline_pair.py
shadeMe: See above.

julian-risch (approved):
Looking forward to seeing this in action in RAGEvaluationHarness!

PR #9, opened 2024-05-28
feat: Add a new `PipelineWrapper` component

masci:
### Related Issues

- fixes #issue-number

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/wrappers/__init__.py
+++ b/haystack_experimental/components/wrappers/__init__.py
@@ -1,0 +1,3 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
--- a/haystack_experimental/components/wrappers/pipeline.py
+++ b/haystack_experimental/components/wrappers/pipeline.py
@@ -9,49 +9,50 @@ from haystack import Pipeline, component
 
 @component
 class PipelineWrapper:
+    """
+    PipelineWrapper wraps a pipeline into a single component.
+
+    This component has the same inputs as the wrapped pipeline. The wrapped pipeline might have
+    a component expecting multiple inputs like this:
+
+    ```python
+    {
+        'llm': {
+            'prompt': {'type': ..., 'is_mandatory': True},
+            'generation_kwargs': {'type': ..., 'is_mandatory': False, 'default_value': None}
+        }
+    }
+    ```
+
+    In turn, this wrapper components would have nested inputs:
+
+    ```python
+    {
+        "this_component": {
+        'llm': {
+            'prompt': {'type': ..., 'is_mandatory': True},
+            'generation_kwargs': {'type': ..., 'is_mandatory': False, 'default_value': None}
+        }
+        }
+    }
+    ````
+
+    This component would be difficult to connect, so we wrap the data we send to the pipeline and
+    the return value of the pipeline using this convention:
+
+    <this component input> -> <wrapped_component_name>:<wrapped_input_name>
+
+    the inputs of this component would then be:
+
+    ```python
+    {
+        'llm:prompt': {...},
+        'llm.generation_kwargs': {...}
+    }
+    ```
+    """
+
     def __init__(self, pipeline: Pipeline) -> None:
-        """
-        PipelineWrapper wraps a pipeline into a single component.
-
-        This component has the same inputs as the wrapped pipeline. The wrapped pipeline might have
-        a component expecting multiple inputs like this:
-
-        ```python
-        {
-            'llm': {
-                'prompt': {'type': ..., 'is_mandatory': True},
-                'generation_kwargs': {'type': ..., 'is_mandatory': False, 'default_value': None}
-            }
-        }
-        ```
-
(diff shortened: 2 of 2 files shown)


mikebellerU:
Hi @masci -- i tried it out but am confused about how it would work.  Here I have an example code -- I create two WrappedPipelines and try to connect them together. 

```
from haystack import Pipeline
from haystack_experimental.components.wrappers.pipeline import PipelineWrapper
from haystack.components.converters import OutputAdapter

p1 = Pipeline()
p1.add_component("adap", OutputAdapter(
    template="Hello {{inp}}", output_type=str))
p2 = Pipeline()
p2.add_component("adap", OutputAdapter(
    template="Goodbye {{inp}}", output_type=str))
p = Pipeline()
p.add_component("pipeline1", PipelineWrapper(p1))
p.add_component("pipeline2", PipelineWrapper(p2))
p.connect("pipeline1.adap:output", "pipeline2.adap:inp")
print(p.run(data={"pipeline1": {"adap:inp": "Paris"}}))
```

 I expect the output to be: `{'pipeline2': {'adap': {'output': 'Goodbye Hello Paris'}}}` but the actual output is: `{'pipeline1': {'adap': {'output': 'Hello Paris'}}}`

Is this because the 'connect' function is not working yet?

But, taking a step back, I also am concerned that this whole concept of SuperPipelines being implemented using Components might get very complicated.  I don't know how a user will keep track of all the levels of data needed. Trying to reuse a Pipeline will require deep knowledge of its internals, which is kind of "code smell".    Thoughts?

masci:
@mikebellerU I fixed the wrap/unwrap of the data passed from the wrapping component to the wrapped pipeline, now the output of your snippet should be `{'pipeline2': {'adap:output': 'Goodbye Hello Paris'}}` as expected.

>Trying to reuse a Pipeline will require deep knowledge of its internals, which is kind of "code smell". Thoughts?

The goal of this component is to treat certain pipelines as "blackboxes" that you can invoke like it was a regular component. I don't think you need to necessarily know the internals of the wrapped pipeline to be able to use it, for example from your snippet if you `print(p.inputs())` you would get `{'pipeline1': {'adap:inp': {'type': typing.Any, 'is_mandatory': True}}}`, which is all you need to use the wrapped pipeline: connect an upstream output with type `Any` to `adap:inp` and you're good to go.

(Any consideration about how complex is passing data to `run()` in Haystack in general, that holds true)

mikebellerU:
Ah!  The patched version works.  I think I also understand your point about input / output.  I really hadn't used these methods enough or really understood their value until you just explained it now -- so thank you for that!  I wonder though whether the PipelineWrapper still isn't quite right.  Here is an example where we I have wrapped a pipeline in a pipeline, in a pipeline.  The pipeline.inputs() call seems to give the correct answer, but when I try to run the pipeline, something is going wrong:

```
from haystack import Pipeline
from haystack_experimental.components.wrappers.pipeline import PipelineWrapper
from haystack.components.converters import OutputAdapter


pg = Pipeline()
pg.add_component("adap", OutputAdapter(
    template="Goodbye {{inp1}}", output_type=str))
print("PG IN:", pg.inputs(), "OUT:", pg.outputs())

ph = Pipeline()
ph.add_component("pg", PipelineWrapper(pg))
ph.add_component("adap", OutputAdapter(
    template="Hello {{inp2}}", output_type=str))
ph.connect("pg.adap:output", "adap.inp2")
print("PH IN:", ph.inputs(), "OUT:", ph.outputs())

pp = Pipeline()
pp.add_component("goodbye", PipelineWrapper(ph))
print("PP IN:", pp.inputs(), "OUT:", pp.outputs())

print(pp.run(data={"goodbye.pg:adap:inp1": "world"}))
```
PP's inputs are described (in the print output) as: `PP IN: {'goodbye': {'pg:adap:inp1': {'type': typing.Any, 'is_mandatory': True}}}`
Even though I believe I have aligned my inputs with this information, this program produces: 

```
Inputs ['goodbye.pg:adap:inp1'] were not matched to any component inputs, please check your run parameters.
Traceback (most recent call last):
  File "/Users/mike/tmp/haystack-experimental/mystuff.py", line 22, in <module>
    print(pp.run(data={"goodbye.pg:adap:inp1": "world"}))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mike/tmp/haystack/.venv/lib/python3.11/site-packages/haystack/core/pipeline/pipeline.py", line 110, in run
    self._validate_input(data)
  File "/Users/mike/tmp/haystack/.venv/lib/python3.11/site-packages/haystack/core/pipeline/base.py", line 638, in _validate_input
    raise ValueError(f"Missing input for component {component_name}: {socket_name}")
ValueError: Missing input for component goodbye: pg:adap:inp1
```

Thoughts?

masci:
After an internal sync we decided not to pursue this component further. The solution doesn't feel quite right and the use cases that would benefit the most (like agentic pipelines) still have unclear requirements.

Currently the problem can be addressed by breaking down pipelines into smaller ones when possible, or considering to have bigger components. 

If anyone from the community wants to take this on, this could make a good integration.

PR #22, opened 2024-06-25
feat: Add API docs generation

vblagoje:
Adds API docs generation for haystack-experimental.


vblagoje:
Let's wait to proceed until @dfokina returns and we have clearer guidelines on the haystack-experimental API documentation. While there's consensus on not providing general component documentation for the experimental package, we need to clarify whether to include API documentation nonetheless. cc @julian-risch @shadeMe

vblagoje:
@shadeMe we agreed yesterday during sprint mid-check to add generated API docs to https://docs.haystack.deepset.ai/reference. 

As experimental classes will change during haystack-experimental lifetime we need to generate these docs continually and frequently for all post 2.3 releases. In other words, haystack-experimental API docs across all haystack releases will always point to the latest haystack-experimental. @dfokina we need your help in setting up this

vblagoje:
@dfokina here is where I need your help the most in this PR. I've used a new category slug called `haystack-experimental` and we want it to show up in the left navigation pane below "Integrations API"
Screenshot 2024-07-01 at 11 57 28

What do we need to do to make this happen?

dfokina:
@vblagoje I will need to manually create a category with the same name in Haystack docs, then the reference will be synced to this slug :) Could we name it `experimental-api` instead, for consistency? Or perhaps `experiments-api` is even better.

vblagoje:
Ok, cool, you have more context and insights here - please name it as you see it to be consistent and logical

dfokina:
Just created the `experiments-api` category :)

.github/workflows/api_docs.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: hatch run readme:sync
+
+      - name: Get version
+        id: version-getter
+        run: |
+          VERSION="$(hatch version | cut -d '.' -f 1,2)"
+          CURRENT_BRANCH="${{ github.ref_name }}"
+          # If we're on `main` branch we should push docs to the unstable version
+          if [ "$CURRENT_BRANCH" = "main" ]; then
+            VERSION="$VERSION-unstable"
```
dfokina, line 51: I am thinking that it's better to do what we did with `core-integrations` API reference, and push the latest docs to all Readme versions simultaneously: https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/workflows/CI_readme_sync.yml since we cannot really version docs out of the main haystack package with the current setup.

docs/README.md
```diff
@@ -0,0 +1,57 @@
+# :ledger: Looking for the docs?
+
+You can find Haystack's documentation at https://docs.haystack.deepset.ai/.
+
+# :computer: How to update docs?
+
+## Overview, Components, Pipeline Nodes, and Guides
```
dfokina, line 7:
```suggestion
## Overview, Components, and Guides
```

docs/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+We use Pydoc-Markdown to create Markdown files from the docstrings in our code. There is a Github Action that regenerates the API pages with each commit.
+
+If you want to generate a new Markdown file for a new Haystack module, create a `.yml` file in `docs/pydoc/config` which configures how Pydoc-Markdown will generate the page and commit it to main.
+
+All the updates to doctrings get pushed to documentation when you commit to the main branch.
+
+### Configuration
+
+Pydoc will read the configuration from a `.yml` file which is located under `/haystack_experimental/docs/pydoc/config. Our files contain three main sections:
```
dfokina, line 33:
```suggestion
Pydoc will read the configuration from a `.yml` file which is located under `/haystack_experimental/docs/pydoc/config`. Our files contain three main sections:
```

dfokina:
@vblagoje left a couple of comments 🔝

.github/workflows/api_docs.yml
vblagoje: Hmm, I think the agreement was to do this for 2.3 and then all subsequent versions. WDYT @shadeMe ?

vblagoje:
And as the last step - we need that `README_API_KEY` set in this project and we should be gtg.

vblagoje:
@shadeMe @dfokina wdyt you approve this PR, we integrate it to main, checks the docs, and we see how to go further from there?

.github/workflows/api_docs.yml
```diff
@@ -0,0 +1,65 @@
...
+  pull_request:
+    paths:
+      - "docs/pydoc/**"
+  push:
+    branches:
+      - main
+      # release branches have the form v1.9.x
+      - "v[0-9]+.[0-9]+.x"
+      # Exclude 1.x release branches, there's another workflow handling those
+      - "!v1.[0-9]+.x"
```
shadeMe, line 13: Looks like a hold over from the core repo?

.github/workflows/api_docs.yml
shadeMe: Since we don't provide compatibility guarantees and currently don't have versioning support for core integrations, I'd go with the core integrations workflow here as well - Only push to readme on release branches.

shadeMe (changes requested):
Could you also add the configs for the eval harness?

vblagoje:
@shadeMe I had to rebase and push here to address that mispelling.

vblagoje:
Any ideas why sync fails now? It works for me locally when I run it with `hatch run readme:sync` from root of the project

dfokina:
@vblagoje I wouldn't go this far, because versions roll out gradually in Readme, so it'll become unnecessarily complex to remember to push to other versions as they come out. I vote for only keeping the latest docs in the docs.

shadeMe:
> @shadeMe I've thought a bit more about our approach given the independent release cycles of haystack and haystack-experimental. To manage compatibility effectively, we should eventually maintain a compatibility matrix within this project. Here's an example:
> haystack Version 	Compatible haystack-experimental Versions
> 2.3 	1.0 - 1.4
> 2.4 	1.8 - 1.x
> 2.5 	1.8 - 2.x
> 
> So that tomorrow when we push a new release of haystack-experimental in pypi we update docs potentially across more than one version of haystack docs. Say we are just about to release haystack-experimental 1.9. We need to update docs for both 2.4 and 2.5 in that case.
> 
> Am I overcomplicating this?

That would indeed be overcomplicating it - The only compatibility assurance that we provide for this package is that it works with the latest public `haystack-ai` release. Let's just keep the latest docs on the website.

vblagoje:
Ok, I agree - so this means we push to latest minor public release and unstable and as we roll releases forward we then remove them from the non-latest release where we previously pushed e.g. we would remove them from 2.2 once 2.3 is released.

vblagoje:
@dfokina @shadeMe @silvanocerza note the latest CI run after the last commit. We run doc sync for the two latest release:
- latest release (i.e. [2.2](https://docs.haystack.deepset.ai/v2.2) now)
- latest unstable (i.e. [2.3-unstable](https://docs.haystack.deepset.ai/v2.3-unstable) now)

That's what we want. Just need to hook them into pypi release job now.
Delete we don't have to solve today, @dfokina tells me it needs an update anyway

docs/pydoc/config/harness.yml
```diff
@@ -0,0 +1,32 @@
+loaders:
+  - type: haystack_pydoc_tools.loaders.CustomPythonLoader
+    search_path: [../../../haystack_experimental/evaluation]
+    modules: ["harness/evaluation_harness",
+              "harness/rag/harness",
+              "harness/rag/evaluation_pipeline",
+              "harness/rag/parameters",
+              "util/pipeline_pair",
+              "util/helpers"]
```
shadeMe, line 9:
```suggestion
 ```
These two modules are internal.

File: docs/pydoc/config/harness.yml [hunk 6]
```diff
@@ -0,0 +1,32 @@
+loaders:
+  - type: haystack_pydoc_tools.loaders.CustomPythonLoader
+    search_path: [../../../haystack_experimental/evaluation]
+    modules: ["harness/evaluation_harness",
+              "harness/rag/harness",
+              "harness/rag/evaluation_pipeline",
```
shadeMe, line 6:
```suggestion
```
This one too.

shadeMe (changes requested):

shadeMe (approved):
LGTM. I'll let @silvanocerza make the final call.

.github/workflows/api_docs.yml
```diff
@@ -0,0 +1,66 @@
...
+  get-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.version_finder.outputs.versions }}
+    steps:
+      - name: Get Haystack Docs versions
+        id: version_finder
+        run: |
+          curl -s "https://dash.readme.com/api/v1/version" --header "authorization: Basic ${{ secrets.README_API_KEY }}" > out
+          VERSIONS=$(jq -c '[ .[] | select(.version | startswith("2.")) | .version ] | .[-2:]' out)
```
silvanocerza, line 25: I would keep the same behaviour we have for the integrations and update all the `2.x` versions.

```suggestion
          VERSIONS=$(jq '[ .[] | select(.version | startswith("2."))| .version ]' out)
```

silvanocerza:
Looks good, I would just change the `api_docs.yml` workflow to sync all 2.x version. Feel free to merge after that change.

vblagoje:
I wrote to Madeesh privately to ask for final advice but it is better to sync all of here.

As you can see - I made one last change to parametrize which versions of docs are we going to sync so we can easily change that number. Normally that would be the last two version 2.x and 2.(x+1)-unstable. We actually provide no guarantees that haystack-experimental still works with some old 2.x version. Now, to test this doc creation I'll change this parameter to 1. We integrate, see how docs looks like - they'll be only added to https://docs.haystack.deepset.ai/v2.3-unstable and then once we release 2.3 I'll revert SYNC_LAST_N_HAYSTACK_VERSIONS to 2 so we can publish updates to 2.3 and 2.4 unstable. At that time I'll also update this workflow to be published only on pypi publish.

I think Silvano said [this](https://github.com/deepset-ai/haystack-experimental/pull/22#issuecomment-2205977553) not having all the context of these docs discussions and it doesn't make sense to update docs to versions we provide no guarantees that experimental works for. 

We also need to add delete docs for version that we are not syncing. But one step at a time.









NEW

[10:15](https://deepset-ai.slack.com/archives/D06F6Q07ZDL/p1720080905156609)
Oh I forgot, we also need to add delete docs for version that we are not syncing. But one step at a time

PR #23, opened 2024-06-25
chore: Move OpenAPI test dependencies to hatch test

vblagoje:
- Moved OpenAPI test dependencies from hatch `envs.default` to `envs.test`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
   "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
-dependencies = ["jsonref", "haystack-ai"]
+dependencies = ["haystack-ai"]
 [project.urls]
 "CI: GitHub" = "https://github.com/deepset-ai/haystack-experimental/actions"
 "GitHub: issues" = "https://github.com/deepset-ai/haystack-experimental/issues"
@@ -48,6 +48,8 @@ dependencies = [
 extra-dependencies = [
   # RAG evaluation harness (SAS evaluator)
   "sentence-transformers>=2.2.0",
+  # OpenAPI dependencies
+  "jsonref",
   # OpenAPI tests
   "cohere",
   "anthropic",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,8 +51,8 @@ extra-dependencies = [
   # OpenAPI dependencies
   "jsonref",
   # OpenAPI tests
-  "cohere",
-  "anthropic",
+  "cohere-haystack",
+  "anthropic-haystack",
   "fastapi",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,9 +39,6 @@ dependencies = [
   # Test
   "pytest",
   "pytest-cov",
-  "fastapi",
-  "cohere",
-  "anthropic",
   # Linting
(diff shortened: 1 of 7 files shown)


shadeMe (commented):
I think `jsonref` is already a dependency of the core package?

silvanocerza:
It's lazily imported in [`openapi_functions.py`](https://github.com/deepset-ai/haystack/blob/fd1a06d17173e7bd5488c1540f6a5dd70a14486a/haystack/components/converters/openapi_functions.py#L18-L19), shall we maybe do the same?

vblagoje:
@silvanocerza this package is tiny but really important for yaml/json spec expansion. It would ease DX to have it automatically included. We have bunch of import guards in OpenAPITool already, having yet another one would be both burden for us and for users.

vblagoje:
> This highlights a problem though.
> 
> Since we have `haystack-experimental` as a dependency in `haystack` we might bring in dependencies that are lazily imported in core but explicit in experimental. This could hide from the our users the need to explicitly specify some dependencies from their projects. And if we update experimental by removing a certain dependency we might break some user's environment. 😕

Ok, good perspective. Is there some middle ground considering that we'll deprecate `OpenAPIServiceConnector` and `OpenAPIServiceToFunctions`? Perhaps we can introduce jsonref in haystack and remove the guards for jsonref in `OpenAPIServiceConnector` and `OpenAPIServiceToFunctions` while we mark them deprecated - all in one shot for 2.3. WDYT @shadeMe

shadeMe:
> Ok, good perspective. Is there some middle ground considering that we'll deprecate `OpenAPIServiceConnector` and `OpenAPIServiceToFunctions`? Perhaps we can introduce jsonref in haystack and remove the guards for jsonref in `OpenAPIServiceConnector` and `OpenAPIServiceToFunctions` while we mark them deprecated - all in one shot for 2.3. WDYT @shadeMe

We shouldn't deprecate/change anything in core until the new component makes it into the former. In other words, `haystack-experimental` should not cause side-effects in `haystack-ai`.

Let's do the following instead:
- Move `jsonref` into a lazy import.
- Add a new `Dependencies` column to the table in the readme that specifies each addition's dependencies, if any.

This should give prospective users an overview of what they should install in their venv without pulling any mandatory dependencies into the experimental package. Given that we aren't planning to release API docs to begin with, I don't see this extra step moving the needle when it come to DX.

vblagoje:
Ok @shadeMe I'll do that in this PR.

pyproject.toml
```diff
@@ -51,6 +48,12 @@ dependencies = [
 extra-dependencies = [
   # RAG evaluation harness (SAS evaluator)
   "sentence-transformers>=2.2.0",
+  # OpenAPI dependencies
+  "jsonref",
+  # OpenAPI tests
+  "cohere",
+  "anthropic",
```
shadeMe, line 55: Don't we also need the Haystack integrations for testing the component itself?

shadeMe (changes requested):

pyproject.toml
vblagoje: Yes, good catch - the integration tests didn't run on CI because we need to add keys first. Let me send you keys first

shadeMe (approved):
Looks good to me. @silvanocerza?

silvanocerza (approved):
Good for me. 👍

PR #25, opened 2024-06-28
chore: Add pytest-rerunfailures, use it for integration tests

vblagoje:
- Improve robustness for integration tests by using pytest-rerunfailures plugin for pytest
- Only add to integration tests, not unit

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dependencies = [
   "mypy",
   # Test
   "pytest",
+  "pytest-rerunfailures",
   "pytest-cov",
   # Linting
   "pylint",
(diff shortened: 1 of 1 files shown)


anakin87:
Are you sure we need to apply this to every integration test, as in core-integrations?

Marking the single tests as flaky could be another option, as we do in Haystack (https://github.com/deepset-ai/haystack/blob/91f57015c0aaccc429cfaaec75eac75a36e36c19/test/core/pipeline/test_draw.py#L15).

vblagoje:
@anakin87 @shadeMe be more reliable now. If we get failures I'll turn of firecrawl openapi integration tests. It is a popular service right now and perhaps sometimes they have scalability issues to serve all the requests.

vblagoje:
> Are you sure we need to apply this to every integration test, as in core-integrations?
> 
> Marking the single tests as flaky could be another option, as we do in Haystack (https://github.com/deepset-ai/haystack/blob/91f57015c0aaccc429cfaaec75eac75a36e36c19/test/core/pipeline/test_draw.py#L15).

Yeah, either way should be ok. The current approach is more "use all the artillery" generic approach as we have bunch off interactions with various services in openapi integration tests.

anakin87 (approved):

PR #41, opened 2024-07-22
fix: Sanitize function names and trim descriptions for LLM compatibility

vblagoje:
#### Why:
Fixes subtle naming bugs to meet LLM naming requirements for function naming patterns and description length limitations. 
Not all operation names from OpenAPI spec and long descriptions are accepted by LLMs.

#### What:
- Implemented function name sanitization to match pattern `^[a-zA-Z0-9_-]+$`
- Trimmed descriptions to max 1024 characters
- Updated schema conversion and operation finding logic
- Added and expanded tests

#### How:
```python
sanitized_name = sanitize_function_name("function-name/with/special_chars")
# Result: "function_name_with_special_chars"

sanitized_desc = sanitize({"description": "Long text..."})
# Result: {"description": "Trimmed to 1024 characters..."}
```

#### How did you test it:
- Unit tests for sanitization functions
- Updated integration tests for edge cases
- Verified compatibility with existing schemas

#### Notes for the reviewer:
- Regex pattern alignment with LLM conventions
- Consistency of sanitized names across the codebase

### DIFF
--- a/haystack_experimental/components/tools/openapi/types.py
+++ b/haystack_experimental/components/tools/openapi/types.py
@@ -41,7 +41,7 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
     return sanitize_function_name(path + "_" + http_method.lower())
 
 
-def sanitize(data: Dict[str, Any]) -> Dict[str, Any]:
+def sanitize_function_calling_definition(data: Dict[str, Any]) -> Dict[str, Any]:
     """
     Sanitizes the given function calling definition by adjusting its properties to LLM requirements.
 
@@ -56,7 +56,7 @@ def sanitize(data: Dict[str, Any]) -> Dict[str, Any]:
             elif key == "description" and isinstance(value, str):
                 sanitized_data[key] = value[:1024]
             else:
-                sanitized_data[key] = sanitize(value)
+                sanitized_data[key] = sanitize_function_calling_definition(value)
         return sanitized_data
     else:
         return data
--- a/haystack_experimental/components/tools/openapi/types.py
+++ b/haystack_experimental/components/tools/openapi/types.py
@@ -41,15 +41,15 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
     return sanitize_function_name(path + "_" + http_method.lower())
 
 
-def sanitize(data: Dict[str, Any]) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
+def sanitize(data: Dict[str, Any]) -> Dict[str, Any]:
     """
-    Sanitizes the given function calling definition(s) by adjusting its properties to LLM requirements.
+    Sanitizes the given function calling definition by adjusting its properties to LLM requirements.
 
     :param data: The function calling definition(s) to sanitize.
     :return: A sanitized function calling definition.
     """
     if isinstance(data, dict):
-        sanitized_data = {}
+        sanitized_data: Dict[str, Any] = {}
         for key, value in data.items():
             if key == "name" and isinstance(value, str):
                 sanitized_data[key] = sanitize_function_name(value)
@@ -58,8 +58,6 @@ def sanitize(data: Dict[str, Any]) -> Union[List[Dict[str, Any]], Dict[str, Any]
             else:
                 sanitized_data[key] = sanitize(value)
         return sanitized_data
-    elif isinstance(data, list):
-        return [sanitize(item) for item in data]
     else:
         return data
 
--- a/haystack_experimental/components/tools/openapi/types.py
+++ b/haystack_experimental/components/tools/openapi/types.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import json
+import re
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
     """
-    if http_method.lower() not in VALID_HTTP_METHODS:
-        raise ValueError(f"Invalid HTTP method: {http_method}")
-    return path.replace("/", "_").lstrip("_").rstrip("_") + "_" + http_method.lower()
+    return sanitize_function_name(path + "_" + http_method.lower())
+
+
+def sanitize(data: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Sanitizes the given function calling definition by adjusting its properties to LLM requirements.
```
shadeMe, line 46: `LLM requirements` is rather vague. Is there some documentation that defines what that means, i.e., where are these requirements coming from?

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
+
+    :param data: The function calling definition(s) to sanitize.
+    :return: A sanitized function calling definition.
+    """
+    if isinstance(data, dict):
```
shadeMe, line 51: Why is this check required?

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
+        sanitized_data: Dict[str, Any] = {}
+        for key, value in data.items():
+            if key == "name" and isinstance(value, str):
```
shadeMe, line 54: Why is the type check needed here? What are the expected types of the value?

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
+                sanitized_data[key] = sanitize_function_name(value)
+            elif key == "description" and isinstance(value, str):
```
shadeMe, line 56: Same as above.

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
     :param http_method: The HTTP method to use for the operationId.
     :returns: The operationId.
     """
-    if http_method.lower() not in VALID_HTTP_METHODS:
-        raise ValueError(f"Invalid HTTP method: {http_method}")
-    return path.replace("/", "_").lstrip("_").rstrip("_") + "_" + http_method.lower()
+    return sanitize_function_name(path + "_" + http_method.lower())
+
+
+def sanitize(data: Dict[str, Any]) -> Dict[str, Any]:
```
shadeMe, line 44: Let's go with a more explicit name - `sanitize_function_calling_definition`?

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
+            elif key == "description" and isinstance(value, str):
+                sanitized_data[key] = value[:1024]
+            else:
+                sanitized_data[key] = sanitize(value)
+        return sanitized_data
+    else:
+        return data
+
+
+def sanitize_function_name(name: str) -> str:
```
shadeMe, line 65: More explicit naming.

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
-    return path.replace("/", "_").lstrip("_").rstrip("_") + "_" + http_method.lower()
+    return sanitize_function_name(path + "_" + http_method.lower())
+
+
+def sanitize(data: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Sanitizes the given function calling definition by adjusting its properties to LLM requirements.
+
+    :param data: The function calling definition(s) to sanitize.
+    :return: A sanitized function calling definition.
```
shadeMe, line 49:
```suggestion
    :returns: A sanitized function calling definition.
```

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -37,9 +38,41 @@ def path_to_operation_id(path: str, http_method: str = "get") -> str:
...
+    else:
+        return data
+
+
+def sanitize_function_name(name: str) -> str:
+    """
+    Sanitizes the function name to match the LLM function naming pattern ^[a-zA-Z0-9_-]+$.
+
+    :param name: The original function name.
+    :return: A sanitized function name that matches the allowed pattern.
```
shadeMe, line 70:
```suggestion
    :returns: A sanitized function name that matches the allowed pattern.
```

shadeMe (changes requested):

haystack_experimental/components/tools/openapi/types.py
vblagoje: @shadeMe I couldn't find this in documentation but you hit it in the error reports returned by LLM once you use enough openapi specs. For example:
```
# BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'tools[0].function.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.", 'type': 'invalid_request_error', 'param': 'tools[0].function.name', 'code': 'invalid_value'}}
```

Once you apply this pattern to method is start to work. Same for length of description field

haystack_experimental/components/tools/openapi/types.py
vblagoje: It is required for recursive search of the function calling dict structure. We might have multiple nested description keys

haystack_experimental/components/tools/openapi/types.py
vblagoje: Yes exactly

haystack_experimental/components/tools/openapi/types.py
shadeMe: Which LLM? Which provider? Are there providers/models that don't accept underscores? If this code doesn't follow an official spec, that we should not make any assumptions and perform this sanitization (normalization, actually) on a per-LLM provider basis at the least.

Also, this kind of impl. detail of LLM-specific requirements shouldn't leak into code that handles OpenAPI - one should perform this normalization just-in-time, i.e, just before passing it to the generator.

haystack_experimental/components/tools/openapi/types.py
shadeMe: Can we formally codify this in some manner? I feel like we have far too many `Any` type hints w.r.t the OpenAPI specific code, and having some kind of schema (or at least better type annotation) would help make this more readable.

haystack_experimental/components/tools/openapi/types.py
vblagoje: Ok, I'll close this PR and rethink it now

vblagoje:
I'll rethink how to do this in a better way outside of OpenAPISpec across LLMs so we can reuse it in the future. Closing

haystack_experimental/components/tools/openapi/types.py
vblagoje: I understand the sentiment @shadeMe . We have nicely, as much as needed imho, typed OpenAPISpec itself. Here we are dealing with these function definitions specified by various LLM providers. They look like this:
- Cohere https://docs.cohere.com/docs/tool-use#step-1
- Anthropic https://docs.anthropic.com/en/docs/build-with-claude/tool-use#tool-use-examples
- OpenAI https://cookbook.openai.com/examples/how_to_call_functions_with_chat_models#basic-concepts

In other words they are just dictionaries. And although it would be nice to bring some order and typing perhaps it is not worth it given how they change and get expanded with new options. Maybe just providing references to these dictionary structures would help a reader quite a lot.

PR #46, opened 2024-07-23
fix: Add tools definitions normalization

vblagoje:
### Why:
This PR introduces necessary normalizations for tool calling definitions because LLM providers (OpenAI, Anthropic, Cohere) have specific requirements for tool definition names and formats. By normalizing tool names and truncating overly long descriptions, we prevent subtle errors when using OpenAPITool and our future tools with various LLMs. 

### What:
- **Utility Functions:** Added `normalize_tool_definition` and `normalize_function_name` for consistent handling of tool definitions and operation IDs. These utility methods have been added (and isolated) in a new module `tools/utils.py` so they can be used in the future by all our LLM invoked tools.
- **Testing:** Added unit tests to cover normalization both as raw unit test and integration test within OpenAPITool context

### How can it be used:
- Usage is internal and not user facing.

### How did you test it:
- Unit tests for utility functions with various input patterns.
- Integration tests with OpenAI, Anthropic, and Cohere OpenAPI spec with intentionally non-conforming names and descriptions

### Notes for the reviewer:
- Verify utility functions are properly introduced
- Note how utility functions are used


haystack_experimental/components/tools/openapi/_openapi.py
```diff
@@ -169,11 +170,32 @@ def get_auth_function(self) -> Callable[[Dict[str, Any], Dict[str, Any]], Any]:
             )
         raise ValueError(f"Unsupported credentials type: {type(self.credentials)}")
 
-    def get_tools_definitions(self) -> List[Dict[str, Any]]:
+    def get_tools_definitions(
+            self,
+            normalization_fn: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None
```
shadeMe, line 175: I don't see any callsites where this parameter is actually used? I'd remove it and use the default normalization function. All of this is internal, so it can be reintroduced if we end up needing the extra flexibility.

Besides, given how most LLM providers take after OpenAI w.r.t to API constraints, we can probably get away with just one normalization function.

haystack_experimental/components/tools/openapi/types.py
```diff
@@ -222,14 +222,26 @@ def from_url(cls, url: str) -> "OpenAPISpecification":
             ) from e
         return cls.from_str(content)
 
-    def find_operation_by_id(self, op_id: str) -> Operation:
+    def find_operation_by_id(
+            self,
+            op_id: str,
+            op_id_normalization_fn: Optional[Callable[[str], str]] = None
```
shadeMe, line 228: Same as above.

shadeMe (changes requested):

shadeMe (approved):

PR #52, opened 2024-08-01
feat: update `ChatMessage` to support tools

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/8136

### Proposed Changes:
I'm overriding the whole module. However, I'm trying not to deviate much from the current implementation...

Notable changes:
- introduce `ToolCall` dataclass to represent a Tool call prepared by the model
- introduce `tool` role for messages containing the results of (external) tool invocations
- drop the support for `function` role (deprecated by OpenAI): I think supporting it properly is not worth it (the world is moving away from this style/mode of doing things). In truth, I would also deprecate it in the Haystack core.

### How did you test it?
Adapted old tests, new tests.

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/dataclasses/chat_message.py
+++ b/haystack_experimental/dataclasses/chat_message.py
@@ -19,16 +19,19 @@ class ChatRole(str, Enum):
 @dataclass
 class ToolCall:
     """
-    Represents a Tool call prepared by the model, usually wrapped in an assistant message.
+    Represents a tool call prepared by the model.
 
-    :param id: The ID of the Tool call.
-    :param tool_name: The name of the Tool to call.
-    :param arguments: The arguments to call the Tool with.
+    It is usually stored in the `tool_calls` attribute of a message from the assistant.
+
+    :param tool_name: The name of the tool to be invoked.
+    :param arguments: The arguments to pass to the tool.
+    :param id: A unique identifier for the tool call. Some providers, such as OpenAI, generate this ID to associate
+        subsequent tool messages to the corresponding tool calls.
     """
 
-    id: str  # noqa: A003
     tool_name: str
     arguments: Dict[str, Any]
+    id: Optional[str]  # noqa: A003
 
 
 @dataclass
@@ -38,8 +41,8 @@ class ChatMessage:
 
     :param content: The text content of the message.
     :param role: The role of the entity sending the message.
-    :tool_call_id: The ID of the tool call that this message is responding to. Applies only to messages from tools.
-    :tool_calls: List of prepared tool calls. Applies only to messages from the assistant.
+    :tool_call_id: The ID of the tool call that a message from a tool is responding to (for messages from tools only).
+    :tool_calls: List of tool calls prepared by the model (for messages from the assistant only).
     :param meta: Additional metadata associated with the message.
     """
 
@@ -66,7 +69,7 @@ class ChatMessage:
         Create a message from the assistant.
 
         :param content: The text content of the message.
-        :param tool_calls: List of prepared tool calls.
+        :param tool_calls: List of tool calls prepared by the model.
         :param meta: Additional metadata associated with the message.
         :returns: A new ChatMessage instance.
         """
@@ -95,12 +98,12 @@ class ChatMessage:
         return cls(content=content, role=ChatRole.SYSTEM, tool_call_id=None, tool_calls=[], meta={})
 
     @classmethod
-    def from_tool(cls, content: str, tool_call_id: str) -> "ChatMessage":
+    def from_tool(cls, content: str, tool_call_id: Optional[str] = None) -> "ChatMessage":
         """
         Create a message from a tool.
 
         :param content: Content of the tool message.
-        :param tool_call_id: Tool call that this message is responding to.
+        :param tool_call_id: The ID of the tool call this message is responding to.
         :returns: A new ChatMessage instance.
         """
         return cls(content=content, role=ChatRole.TOOL, tool_call_id=tool_call_id, tool_calls=[], meta={})
--- a/haystack_experimental/dataclasses/chat_message.py
+++ b/haystack_experimental/dataclasses/chat_message.py
@@ -4,7 +4,7 @@
 
 from dataclasses import asdict, dataclass, field
 from enum import Enum
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Sequence, Union
 
 
 class ChatRole(str, Enum):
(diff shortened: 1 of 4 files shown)


haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,136 @@
...
+    Represents a Tool call prepared by the model, usually wrapped in an assistant message.
+
+    :param id: The ID of the Tool call.
+    :param tool_name: The name of the Tool to call.
+    :param arguments: The arguments to call the Tool with.
+    """
+
+    id: str  # noqa: A003
+    tool_name: str
+    arguments: Dict[str, Any]
```
anakin87, line 31: `Dict[str, Any]` is very generic. I propose to iterate over it as the experiment proceeds.
Some providers return a JSON, while OpenAI returns a JSON dumped string...

anakin87:
I explored various approaches to improve the abstraction for `ChatMessage`.


1. One attempt involved creating a class hierarchy.
```python
@dataclass
class BaseChatMessage():
    content: str
    role: ChatRole
    meta: Dict[str, Any] = field(default_factory=dict, hash=False)

...


@dataclass
class UserChatMessage(BaseChatMessage):
    def __init__(self, content: str):
        super().__init__(content=content, role=ChatRole.USER)


@dataclass
class AssistantChatMessage(BaseChatMessage):
    tool_calls: List[ToolCall] = field(default_factory=list)

    def __init__(self, content: str, tool_calls: Optional[List[ToolCall]] = None, meta: Optional[Dict[str, Any]] = None):
        super().__init__(content=content, role=ChatRole.ASSISTANT, meta=meta or {})
        self.tool_calls = tool_calls or []
...        
```
However, I was not satisfied with the result and felt uncertain about the overall design.
This design would be breaking, without introducing advantages over the current use of class methods for creating messages.


2. Another attempt involved making the `content` more generic:
```python
@dataclass
class TextPart:
    content: str

@dataclass
class ToolPart:
    tool_calls: List[ToolCall]    

@dataclass
class ChatMessage:
    """
    Represents a message in a LLM chat conversation.

    :param content: The text content of the message.
    :param role: The role of the entity sending the message.
    :param meta: Additional metadata associated with the message.
    """

    content: Union[List[Union[TextPart, ToolPart]], str]
    role: ChatRole
    tool_call_id: Optional[str] = None
    meta: Dict[str, Any] = field(default_factory=dict, hash=False)
```

While this initially made sense to me, it resulted in a worse user experience.
Our users frequently interact directly with `ChatMessage`. It is not just an internal abstraction.
While for message creation, we could provide shortcuts for handling `str`, this can't be easily done for the output coming from a LLM.


Given this,  I have found that our current abstraction is not bad and provides a good grade of flexibility, while being sufficiently usable.
So I'm sticking to it in this PR.

anakin87:
I took the approach we agreed on, similar to the second one in the previous message, with some adjustments that improve the UX.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+from typing import Any, Dict, List, Optional, Sequence, Union
+
+
+class ChatRole(str, Enum):
+    """Enumeration representing the roles within a chat."""
+
+    ASSISTANT = "assistant"
+    USER = "user"
+    SYSTEM = "system"
+    TOOL = "tool"
```
shadeMe, line 16: Docstrings for each variant describing what the chatmessage would look like for each of them.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+class ToolCallResult:
+    """
+    Represents the result of a Tool invocation.
+
+    :param result: The result of the Tool invocation.
+    :param origin: The Tool call that produced this result.
+    """
+
+    result: str
+    origin: Optional[ToolCall] = None
```
shadeMe, line 44: Why is this optional? There should always be an originating toolcall for every result?

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+ChatMessageContentT = Union[TextContent, ToolCall, ToolCallResult]
+
+
+@dataclass
+class ChatMessage:
+    """
+    Represents a message in a LLM chat conversation.
+
+    :param content: The content of the message.
+        It is a list containing one or more of the following types: `TextContent`, `ToolCall`, `ToolCallResult`.
```
shadeMe, line 67:
```suggestion
```
This bit should be self-evident from the type annotations (let's make sure that we export `chatMessageContentT` to the API docs).

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+        content: List[ChatMessageContentT] = []
+        if text:
+            content.append(TextContent(text=text))
+        if tool_calls:
+            content.extend(tool_calls)
+
+        return cls(_role=ChatRole.ASSISTANT, _content=content, _meta=meta or {})
+
+    @classmethod
+    def from_tool(cls, result: str, origin: Optional[ToolCall] = None) -> "ChatMessage":
```
shadeMe, line 197: Origin should be non-optional. Also, let's rename `result` to `tool_result`.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+        return cls(_role=ChatRole.TOOL, _content=[ToolCallResult(result=result, origin=origin)])
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Converts ChatMessage into a dictionary.
+
+        :returns:
+            Serialized version of the object.
+        """
+        serialized: dict = {}
```
shadeMe, line 214:
```suggestion
        serialized: Dict[str, Any] = {}
```

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+        serialized["_role"] = self._role.value
+        serialized["_meta"] = self._meta
+
+        content: List[dict] = []
```
shadeMe, line 218:
```suggestion
        content: List[Dict[str, Any]] = []
```

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,254 @@
...
+        for part in self._content:
+            if isinstance(part, TextContent):
+                content.append({"text": part.text})
+            elif isinstance(part, ToolCall):
+                content.append({"tool_call": asdict(part)})
+            elif isinstance(part, ToolCallResult):
```
shadeMe, line 224: Let's add an `else` clause that raises as a sanity check to both serde methods.

test/dataclasses/test_chat_message.py
```diff
@@ -0,0 +1,166 @@
...
+
+    serialized_message = message.to_dict()
+
+    assert serialized_message == {
+        "_content": [{"text": "a message"}, {"tool_call": {"id": "123", "tool_name": "tool", "arguments": {"a": 1}}}],
+        "_role": "assistant",
+        "_meta": {},
+    }
+
+    assert ChatMessage.from_dict(serialized_message) == message
```
shadeMe, line 166: Let's extend these tests to account for all content types, i.e., let's add each of the three supported payloads to the message and then perform to/from dict, etc. On the same note, we should also test for the exception handling (as suggested by the above comment).

shadeMe (changes requested):
Just a few minor changes 🚀

haystack_experimental/dataclasses/chat_message.py
anakin87: Let's make this mandatory.
We can change that in the future if for some reason we find it limits us too much.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -8,11 +8,19 @@
...
 class ChatRole(str, Enum):
-    """Enumeration representing the roles within a chat."""
+    """
+    Enumeration representing the roles within a chat.
+
+    - `USER`: The user role. A message from the user contains only text.
+    - `SYSTEM`: The system role. A message from the system contains only text.
+    - `ASSISTANT`: The assistant role. A message from the assistant can contain text and prepared Tool calls.
+        It can also store additional metadata.
+    - `TOOL`: The tool role. A message from a tool contains the result of a Tool invocation.
```
shadeMe, line 18: Let's move this to each variant:
```python
#: The user role. A message from the user contains only text.
USER = "user"
...
```

shadeMe (approved):

Gostnpc1 (commented):
Nice

PR #53, opened 2024-08-07
feat: Tool class

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/8166

### Proposed Changes:
Introduce a new `Tool` dataclass, that can store a tool specification and allow the invocation of the tool.

### How did you test it?
CI, new tests.

### Notes for the reviewer
- minimal: the Language Model only needs the tool spec; `invoke` is added to handle tool invocation in our framework
- in the future, we will introduce other more convenient/automated ways of creating a tool: starting from a function, etc.

*On typing of parameters*

I tried to introduce something like this
```python
class Property(TypedDict):
  type: str
  description: str
  enum: NotRequired[List[str]]

class Parameters(TypedDict):
  type: str
  required: List[str]
  properties: Dict[str, Property]
```

Unfortunately, mypy complains if you do:
```python
params = {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}

tool = Tool(name="", description="", parameters=params, function=lambda x:x)
```

> Argument "parameters" to "Tool" has incompatible type "dict[str, Collection[str]]"; expected "Parameters"

To fix this, you should explicitly do `params: Parameters = {...}`.

For this reason, in the current PR, I did not introduce typing for parameters, but we can also decide otherwise.
TBH, I want to provide flexibility, but I do not expect many users to manually input the JSON schema, especially if we provide convenient methods.

*On validation*

- I'm only validating the JSON schema: if the schema is invalid, we will get an error from the LM API
- several consistency checks could be done between the function and the schema
- we could validate `kwargs` used to invoke the Tool against the schema

In general, when we will introduce more powerful constructors (e.g.: `@from_function` class method), I would expect most of the users to use them. So it can be ok to provide more freedom to people manually building this object from scratch.

### Checklist

- I documented my code


haystack_experimental/dataclasses/tool.py
```diff
@@ -0,0 +1,93 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import asdict, dataclass
+from typing import Any, Callable, Dict
+
+from haystack.lazy_imports import LazyImport
+from haystack.utils import deserialize_callable, serialize_callable
+
+with LazyImport(message="Run 'pip install jsonschema'") as jsonschema_import:
```
shadeMe, line 11: We should add this dependency to the table in the readme.

haystack_experimental/dataclasses/tool.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Data class representing a tool for which Language Models can prepare a call.
+
+    Accurate definitions of the textual attributes such as `name` and `description`
+    are important for the Language Model to correctly prepare the call.
+
+    :param name: Name of the tool.
+    :param description: Description of the tool.
+    :param parameters: A JSON schema defining the parameters expected by the tool.
+    :param function: The callable that will be invoked when the tool is called.
```
shadeMe, line 35:
```
    :param name: 
		Name of the tool.
```
Let's format these docstring correctly.

shadeMe (approved):

haystack_experimental/dataclasses/tool.py
anakin87: done

haystack_experimental/dataclasses/tool.py
anakin87: done

PR #54, opened 2024-08-08
chore: Disable meteo openapi tests

vblagoje:
Disabling unit test for meteo openapi tests, underlying service often overloaded

PR #55, opened 2024-08-09
feat: Add unstable marker, mark unstable tests, skip them in CI

vblagoje:
- Add new marker for tests `unstable`
- Mark all unstable tests and reasons why they fail
- Skip these tests in CI by default
- Run them from command line if needed with:
    - `hatch run test:integration -- -m "integration and unstable"` 
    
Supersedes https://github.com/deepset-ai/haystack-experimental/pull/54

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,8 +60,8 @@ extra-dependencies = [
 ]
 
 [tool.hatch.envs.test.scripts]
-unit = 'pytest --cov-report xml:coverage.xml --cov="haystack_experimental" -m "not integration and not unstable" {args:test}'
-integration = 'pytest --reruns 3 --reruns-delay 60 -x --maxfail=5 -m "integration and not unstable" {args:test}'
+unit = 'pytest --cov-report xml:coverage.xml --cov="haystack_experimental" -m "not integration" {args:test}'
+integration = 'pytest --reruns 3 --reruns-delay 60 -x --maxfail=5 -m "integration" {args:test}'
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -60,14 +60,14 @@ extra-dependencies = [
 ]
 
 [tool.hatch.envs.test.scripts]
-unit = 'pytest --cov-report xml:coverage.xml --cov="haystack_experimental" -m "not integration" {args:test}'
-integration = 'pytest --reruns 3 --reruns-delay 60 -x --maxfail=5 -m "integration" {args:test}'
+unit = 'pytest --cov-report xml:coverage.xml --cov="haystack_experimental" -m "not integration and not unstable" {args:test}'
+integration = 'pytest --reruns 3 --reruns-delay 60 -x --maxfail=5 -m "integration and not unstable" {args:test}'
```
shadeMe, line 64: Let's not modify the hatch command to directly exclude these tests - this excludes those tests on non-CI hosts as well. Instead, let's pass an argument that skips them and modify the workflow to include that argument.

shadeMe (changes requested):

pyproject.toml
vblagoje: Ok, np, this should be the [fix](https://github.com/deepset-ai/haystack-experimental/pull/55/commits/c636ae823e0795ba354ff85fb9aa093fc617d8ed)

shadeMe (approved):

PR #58, opened 2024-08-14
feat: Tool invoker

anakin87:
fixes https://github.com/deepset-ai/haystack/issues/8187


haystack_experimental/components/tools/toolinvoker.py
```diff
@@ -0,0 +1,54 @@
...
+
+@component
+class ToolInvoker:
+    def __init__(self, tools: List[Tool], raise_on_failure: bool = False):
+        self.tools = tools
+        self._tool_names = {tool.name for tool in tools}
+        self.raise_on_failure = raise_on_failure
+
+    @component.output_types(tool_results=List[ChatMessage])
+    def run(self, tool_message: ChatMessage):
```
anakin87, line 27: Since the input message can contain one or more tool calls, I think it makes sense to return a list of (tool) messages, each of them wrapping a tool result.

haystack_experimental/components/tools/toolinvoker.py
```diff
@@ -0,0 +1,54 @@
...
+                try:
+                    tool_response = tool_to_call.invoke(**tool_arguments)
+                    tool_results.append(ChatMessage.from_tool(tool_result=json.dumps(tool_response), origin=tool_call))
+
+                except ToolInvocationError as e:
+                    if self.raise_on_failure:
+                        raise e
+                    tool_results.append(
+                        ChatMessage.from_tool(tool_result=_FUNCTION_RUN_FAILURE.format(error=e), origin=tool_call)
+                    )
```
anakin87, line 51: I think it makes sense to give users the option not to raise an exception on tool invocation failure, but to return a tool result with an error message. This makes sense to me in a conversational scenario.

We can probably improve this message by changing the text and adding more details about the attempted tool invocation.

haystack_experimental/components/tools/toolinvoker.py
```diff
@@ -0,0 +1,54 @@
...
+            else:
+                raise Exception(f"Tool {tool_name} not found in the list of tools.")
```
anakin87, line 53: In this case, I raise an exception: the latest models do not often hallucinate tool names, so if the tool is not found it probably means that the user forgot to provide it to the tool invoker.

I am not sure if we want to use `raise_on_failure` in this case as well. :thinking:

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -35,9 +35,9 @@ class ToolCall:
     :param arguments: The arguments to call the Tool with.
     """
 
-    id: Optional[str]  # noqa: A003
     tool_name: str
     arguments: Dict[str, Any]
+    id: Optional[str] = None  # noqa: A003
```
anakin87, line 40: unrelated change.

Previously, I forgot to provide a default for `id`.
Dataclasses require attributes with defaults to be placed after others.

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -0,0 +1,204 @@
...
+        :param tools:
+            A list of tools that can be invoked.
+        :param raise_on_failure:
+            If True, the component will raise exceptions in case the tool is not found or invocation fails.
+            If False, the component will return a ChatMessage from the tool, wrapping the error message.
+        """
+        if not tools:
+            raise ValueError("ToolInvoker requires at least one tool to be provided.")
+
+        self.tools = tools
```
shadeMe, line 104: We should ensure that the tools have unique names (just realized that we don't do this in the generator either; let's fix that too).

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -0,0 +1,204 @@
...
+        self._tools_with_names = {tool.name: tool for tool in tools}
+        self.raise_on_failure = raise_on_failure
+
+    @staticmethod
+    def _convert_tool_result_to_string(result: Any) -> str:
```
shadeMe, line 109: Why don't we use `raise_on_failure` here?

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -0,0 +1,204 @@
...
+            return json.dumps(result)
+        except TypeError as e:
+            logger.warning(
+                "Failed to convert tool result to string using `json.dumps`. Error: {error}. Falling back to `str`.",
+                error=e,
+            )
+            return str(result)
+
+    @component.output_types(tool_messages=List[ChatMessage])
+    def run(self, message: ChatMessage):
```
shadeMe, line 128: Why just a single chat message? It'd be more useful to use the same modality of the generator's outputs, i.e., a list of chat messages are accepted and each tool call message is replaced with the tool call result(s).

That way, you don't need another component in-between the generator and the invoker to preprocess the former's outputs.

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -0,0 +1,204 @@
...
+
+        for tool_call in tool_calls:
+            tool_name = tool_call.tool_name
+            tool_arguments = tool_call.arguments
+
+            if not tool_name in self._tools_with_names:
+                msg = _TOOL_NOT_FOUND.format(tool_name=tool_name, available_tools=self._tools_with_names.keys())
+                if self.raise_on_failure:
+                    raise ToolNotFoundException(msg)
+                tool_messages.append(ChatMessage.from_tool(tool_result=msg, origin=tool_call))
```
shadeMe, line 162: I think it would be useful to have a flag in the tool call result payload to indicate errors.

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -0,0 +1,204 @@
...
+
+        :param data:
+            The dictionary to deserialize from.
+        :returns:
+            The deserialized component.
+        """
+        init_params = data.get("init_parameters", {})
+        serialized_tools = init_params.get("tools", [])
+        deserialized_tools = [Tool.from_dict(tool_data) for tool_data in serialized_tools]
+        data["init_parameters"]["tools"] = deserialized_tools
```
shadeMe, line 203: Let's add a utility function similar to `deserialize_secrets_inplace` for tools? I think we'll need it once we start supporting more generators.

test/components/tools/test_tool_invoker.py
```diff
@@ -0,0 +1,195 @@
...
+        data = invoker.to_dict()
+        assert data == {
+            "type": "haystack_experimental.components.tools.tool_invoker.ToolInvoker",
+            "init_parameters": {
+                "tools": [weather_tool.to_dict()],
+                "raise_on_failure": True
+            },
+        }
+
+    def test_from_dict(self, weather_tool):
```
shadeMe, line 184: Test that serializes/deserializes a pipeline with the component.

shadeMe (changes requested):

haystack_experimental/components/tools/tool_invoker.py
anakin87: this was my initial idea.
Then I realized that even if some (good) results are not JSON serializable, their string representation can be OK for humans/language models.
Think for example of a datetime object...

haystack_experimental/components/tools/tool_invoker.py
anakin87: In the most common scenarios, you have a _single_ message from assistant with one or more tool calls and you then execute these invocations, returning a list of (one or more) messages.
I'm quite confident that this design can fit into scenarios like the one outlined by [my old experiments] (https://colab.research.google.com/drive/1ZxUiTsYDBckktywK65wYazEZCqVVyHlA?usp=sharing) with some small changes.

Our current [`OpenAIFunctionCaller`](https://github.com/deepset-ai/haystack-experimental/blob/main/haystack_experimental/components/tools/openai/function_caller.py) takes `List[ChatMessage] ` as input but only processes the first one.

So if we take a list of messages with tool calls as input we can alternatively choose to:
- process the first or the last one
- return a list of list of messages
- (flat the previous list of lists)

Happy to discuss...

haystack_experimental/components/tools/tool_invoker.py
anakin87: I created a separated PR for that: #61
Once the PR is merged, I will rebase this PR and use the utility function

haystack_experimental/components/tools/tool_invoker.py
anakin87: (PR for generator -> #62)

anakin87: done

test/components/tools/test_tool_invoker.py
anakin87: done

haystack_experimental/components/tools/tool_invoker.py
anakin87: makes sense
done!

haystack_experimental/components/tools/tool_invoker.py
anakin87: Now we default to `str` and allow customizability via a dedicated init parameter.

For consistency, in case of errors, with `raise_on_failure=False`, we return a `ChatMessage` containing an error.

haystack_experimental/components/tools/tool_invoker.py
anakin87: We discussed this offline and there could be several different sensible ways to structure inputs and outputs for this component.

During the discussion, I liked the idea of accepting a list of messages, processing only those with tool calls, and returning a list of (tool) messages. So I modified the implementation in this direction.

However, after merging this PR I will play with the current implementation and see if it is ergonomic and fits well with common use cases. We can then modify it later.

haystack_experimental/components/tools/tool_invoker.py
anakin87: done

shadeMe (changes requested):

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -88,92 +100,114 @@ def dummy_weather_function(city: str):
...
+                if self.raise_on_failure:
+                    raise StringConversionError("Failed to convert tool result to string using `str`") from e
+                tool_result_str = _TOOL_RESULT_CONVERSION_FAILURE.format(error=e, conversion_function="str")
+                error = True
+
+        return ChatMessage.from_tool(tool_result=tool_result_str, error=error, origin=tool_call)
 
     @component.output_types(tool_messages=List[ChatMessage])
-    def run(self, message: ChatMessage):
+    def run(self, messages: List[ChatMessage]):
```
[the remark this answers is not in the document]
anakin87: done

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -88,92 +100,114 @@ def dummy_weather_function(city: str):
...
-                    raise e
-                msg = _TOOL_INVOCATION_FAILURE.format(error=e)
-                tool_messages.append(ChatMessage.from_tool(tool_result=msg, origin=tool_call))
-                continue
-
-            tool_result_string = self._convert_tool_result_to_string(tool_response)
-            tool_messages.append(ChatMessage.from_tool(tool_result=tool_result_string, origin=tool_call))
+        for message in messages:
+            tool_calls = message.tool_calls
+            if tool_calls:
```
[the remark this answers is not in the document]
anakin87: right

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -88,92 +100,114 @@ def dummy_weather_function(city: str):
...
 
         self.tools = tools
-        self._tools_with_names = {tool.name: tool for tool in tools}
+        self._tools_with_names = dict(zip(tool_names, tools))
         self.raise_on_failure = raise_on_failure
+        self.convert_result_to_json_string = convert_result_to_json_string
 
-    @staticmethod
-    def _convert_tool_result_to_string(result: Any) -> str:
+    def _prepare_tool_message(self, result: Any, tool_call: ToolCall) -> ChatMessage:
```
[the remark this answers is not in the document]
anakin87: done

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -47,10 +47,12 @@ class ToolCallResult:
 
     :param result: The result of the Tool invocation.
     :param origin: The Tool call that produced this result.
+    :param error: Whether the Tool invocation resulted in an error.
     """
 
     result: str
     origin: ToolCall
+    error: bool = False
```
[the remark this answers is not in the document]
anakin87: done

test/components/tools/test_tool_invoker.py
```diff
@@ -186,10 +248,77 @@ def test_from_dict(self, weather_tool):
...
+            'connections': [
+                {
+                    'sender': 'invoker.tool_messages',
+                    'receiver': 'chatgenerator.messages'
+                }
+            ]
+        }
+
+
+        new_pipeline = Pipeline.from_dict(pipeline_dict)
```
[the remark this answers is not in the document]
anakin87: added

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -257,9 +257,11 @@ def from_dict(cls, data: Dict[str, Any]) -> "ChatMessage":
             elif "tool_call" in part:
                 content.append(ToolCall(**part["tool_call"]))
             elif "tool_call_result" in part:
+                print(part)
```
shadeMe, line 260:
```suggestion
```

shadeMe (approved):
LGTM. One small fix.

PR #59, opened 2024-08-19
fix: `OpenAPITool` - use keyword args for chat generator run invocation

vblagoje:
In the main haystack project pull request [8054](https://github.com/deepset-ai/haystack/pull/8054) added a new parameter in chat generator run methods.

Since we didn't use keyword arguments in invocation of the run method of the chat generator in [`openapi_tool.py`](https://github.com/deepset-ai/haystack-experimental/blob/main/haystack_experimental/components/tools/openapi/openapi_tool.py#L161) we need to now use keyword arguments.


vblagoje:
Madeesh - this fix is blocking @davidsbatista and others from moving forward.

anakin87 (approved):

shadeMe (approved):

PR #65, opened 2024-08-28
feat: Implement `AsyncPipeline`

shadeMe:
### Related Issues

- soft-dependency on https://github.com/deepset-ai/haystack/pull/8279 

### Proposed Changes:

 
 
Introduces the `AsyncPipeline` class that supports `async` Python runtimes. Uses the same execution logic as a regular `Pipeline` with the following changes:
- The `run` method of the `AsyncPipeline` returns an asynchronous generator that yields the outputs of executed components immediately after they are run.
	- This allows for the streaming of intermediate results in an async environment, e.g: a RAG pipeline can yield the results of the retriever before the generator is run.
	- The last output yielded by the `run` method is the pipeline's final output.
	- Since method is a generator, the execution of the pipeline can only progress if its outputs are consumed. Turning that around, it allows the user to determine - after each component is executed - if the pipeline should continue running.
- Components that implement an `async_run` method will be executed concurrently.
- Components that do not implement an `async_run` method will be launched on a thread pool and waited on until they return a result.

In following PRs, we'll introduce variants of components that benefit from async support, i.e., I/O-bound components, primarily generators and document stores (and their associated components).

### How did you test it?

Unit tests

### Notes for the reviewer

The files under `test/core/pipeline/features` can be ignored as they've been directly lifted from the core package.

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/core/__init__.py
+++ b/haystack_experimental/core/__init__.py
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from .pipeline import AsyncPipeline, run_async_pipeline
+
+_all_ = ["AsyncPipeline", "run_async_pipeline"]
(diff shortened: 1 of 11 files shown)


shadeMe:
Test failures are unrelated.

shadeMe:
Blocked from merging until `haystack-experimental 0.2.0`'s release.

PR #84, opened 2024-09-06
fix: Don't run OpenAPITool tests on empty secrets

vblagoje:
Skips OpenAPITool tests when env vars are specified (present in os.environ) but are empty strings (e.g. PR is from fork and secrets don't get set)


shadeMe (approved):
Thanks for the fix!

PR #92, opened 2024-09-13
feat: metadata extractor based on a LLM

davidsbatista:
### Related Issues

- fixes [#5700](https://github.com/deepset-ai/haystack/issues/5700) and [#5702](https://github.com/deepset-ai/haystack/issues/5702)

### Proposed Changes:

- Adding a new component which relies on a LLM and a prompt to extract metadata from a Document. Meant to be used within a pipeline during indexing. This was a request from @sjrl - have a look, you can pull this branch and test it

### How did you test it?

- added  unit tests and integration tests
- manual verification

### Checklist

- I documented my code


julian-risch:
@davidsbatista We should list the new component in the catalog as part of this PR: https://github.com/deepset-ai/haystack-experimental?tab=readme-ov-file#experiments-catalog

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,172 @@
...
+    def __init__(
+            self,
+            prompt: str,
+            expected_keys: List[str],
+            model: Optional[str] = None,
+            raise_on_failure: bool = True
+    ):
+        self.prompt = prompt
+        self.builder = PromptBuilder(prompt)
+        self.generator = OpenAIGenerator() if model is None else OpenAIGenerator(model=model)
```
sjrl, line 88: Instead of just choosing the OpenAIGenerator, could we make the generator an init param? We have situations where need to use other generators for performance reasons or deployment reasons. E.g. We need to use AzureOpenAIGenerator or the AWSBedrockGenerator for enterprise requirements. 

Also the way this is currently specified it's not possible to pass any additional params to the Generator component like the name of the env variable that stores the API key or generation params like setting temp to 0.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,172 @@
...
+       }
+    >>
+    ```
+    """ # noqa: E501
+    def __init__(
+            self,
+            prompt: str,
+            expected_keys: List[str],
+            model: Optional[str] = None,
+            raise_on_failure: bool = True
```
sjrl, line 84: I think these lines are over indented.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,172 @@
...
+            True if the received output is a valid JSON with the expected keys, False otherwise.
+        """
+        try:
+            parsed_output = json.loads(received)
+        except json.JSONDecodeError:
+            msg = "Response from LLM evaluator is not a valid JSON."
+            if self.raise_on_failure:
+                raise ValueError(msg)
+            warn(msg)
+            return False
```
sjrl, line 116: I'm not entirely sure how we could accomplish this, but in the situation that there is a JSONDecodeError it would be amazing if this component could either:
- Recall the Generator with the error message appended to try and fix the issue. 
- If the above is too much scope for a single component, instead provide enough info (e.g. the error message) as part of the output of this component so that other components in the pipeline could be added to fix the issue. 

What do you think?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,172 @@
...
+        :returns:
+            True if the received output is a valid JSON with the expected keys, False otherwise.
+        """
+        try:
+            parsed_output = json.loads(received)
+        except json.JSONDecodeError:
+            msg = "Response from LLM evaluator is not a valid JSON."
+            if self.raise_on_failure:
+                raise ValueError(msg)
+            warn(msg)
```
sjrl, line 115: For this warning could we use the haystack logger?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,172 @@
...
+            return False
+
+        if not all(output in parsed_output for output in expected):
+            msg = f"Expected response from LLM evaluator to be JSON with keys {expected}, got {received}."
+            if self.raise_on_failure:
+                raise ValueError(msg)
+            warn(msg)
```
sjrl, line 122: Same here could we use the Haystack logger to log the warning message?

haystack_experimental/components/extractors/llm_metadata_extractor.py
sjrl: Also could we add docstrings for these? E.g. I think it would be good to be very clear that the variable `input_text` is required in the prompt string otherwise this component won't work.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,172 @@
...
+    """ # noqa: E501
+    def __init__(
+            self,
+            prompt: str,
+            expected_keys: List[str],
+            model: Optional[str] = None,
+            raise_on_failure: bool = True
+    ):
+        self.prompt = prompt
+        self.builder = PromptBuilder(prompt)
```
sjrl, line 87: Could we also set the `required_variables` param here to avoid silent failures in case for some reason the `input_text` isn't passed at run time?
```suggestion
        self.builder = PromptBuilder(prompt, required_variables=["input_text"])
```

sjrl:
Hey @davidsbatista thanks for you work on this! 

One idea that came to mind that this component doesn't yet quite solve is how to handle extracting global information about a Document (at least not elegantly). For example, let's say I want to extract the title from a really long PDF (e.g. more than 200 pages) and store this title in all future chunks I create from this PDF file. With the way this component works I would need to do something like: 

File -> PDFConverter -> LLMMetadataExtractor -> DocumentSplitter -> DocumentWriter

but since we have such a long PDF there is a good chance that it will surpass the context window of the LLM and in reality all I need to do is send the first few pages of the file to find the title so this version also feels inefficient latency and cost-wise. 

So I was wondering if we could brainstorm a way to handle this type of use case as well? Maybe simply to allow a user to specify a page range as well at init time so we don't need to send the whole Document to the LLM?

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: It's not so easy to generalise this to any Generator, since there's no abstraction. I will for now, just add these 3:

- OpenAIGenerator
- AzureOpenAIGenerator
- AmazonBedrockGenerator

any other you would like to have?

sjrl: Just double checking, but generalizing here would just be about having an expanded type for Generator right? I believe all Generators follow the same run method signature or are we worried that isn't respected?

sjrl: Otherwise, I think those three + the GeminiGenerator should be good for now.

davidsbatista: > Just double checking, but generalizing here would just be about having an expanded type for Generator right? I believe all Generators follow the same run method signature or are we worried that isn't respected?

Yes, but that means making this PR dependent on a change on Haystack core - but we can definitely think about it.

> Otherwise, I think those three + the GeminiGenerator should be good for now.
Alright, adding that one as well

davidsbatista:
```
    def __init__(
        self,
        prompt: str,
        input_text: str,
        expected_keys: List[str],
        raise_on_failure: bool = True,
        generator: Union[OpenAIGenerator, AzureOpenAIGenerator, AmazonBedrockGenerator],
        model: Optional[str] = None,
        generation_kwargs: Optional[Dict[str, Any]] = None,
    ):
```

If we have an `__init__()` like this wouldn't make sense that the generator is already initalised and passed to the `LLMMetadataExtractor()`'s `__init()__` and we remove the` model` and `generation_kwargs`.

Alternatively we change the generator to be an `Enum` and depending on it's value we initialised on of the possible generators it inside the `LLMMetadataExtractor`

Do you see any other possibilty?

The whole issue here is wanting to be free to use any possible `Generator` but not having an abstraction to represent it.

sjrl: > If we have an __init__() like this wouldn't make sense that the generator is already initalised and passed to the LLMMetadataExtractor()'s __init()__ and we remove the model and generation_kwargs.

Yeah I think this makes the most sense. I'd prefer this approach.

davidsbatista:
@sjrl still fixing some tests, but please have a look and let me know what you think now.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,220 @@
...
+import json
+import logging
+from typing import Any, Dict, List, Tuple, Union, Type
+from warnings import warn
+
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.components.builders import PromptBuilder
+from haystack.components.generators import AzureOpenAIGenerator, OpenAIGenerator
+from haystack_integrations.components.generators.amazon_bedrock import AmazonBedrockGenerator
+from haystack_integrations.components.generators.google_vertex import VertexAIGeminiGenerator
```
sjrl, line 15: Should we be using Lazy imports here for the integration imports?

davidsbatista: This will just solve the issue about notifying the user that he/she needs to install those packages - for testing, something else needs to be added to the CI, I suspect.

sjrl: Yeah I think warning the user would still be good, but yes a different solution is needed for the CI

dfokina:
@davidsbatista Pls add the component to the pydocs, too :) It should be in a new `extractors_api.yml` file, with `order: 50`

davidsbatista:
@dfokina good point, thanks for remembering me of that

docs/pydoc/config/extractors_api.yml
```diff
@@ -0,0 +1,27 @@
...
+  - type: filter
+    expression:
+    documented_only: true
+    do_not_filter_modules: false
+    skip_empty_modules: true
+  - type: smart
+  - type: crossref
+renderer:
+  type: haystack_pydoc_tools.renderers.ReadmeCoreRenderer
+  excerpt: Extracting information from Documents.
```
dfokina, line 16:
```suggestion
  excerpt: Extracting information from documents.
```

docs/pydoc/config/extractors_api.yml
```diff
@@ -0,0 +1,27 @@
...
+  category_slug: experiments-api
+  title: EXtractors
```
dfokina, line 18:
```suggestion
  title: Extractors
```

dfokina (commented):
Thanks! Just a couple of fixes:

haystack_experimental/components/extractors/llm_metadata_extractor.py
sjrl: Also to quickly pop in, rolling back on this really limits the usability of the component for some clients. I recognize the nesting of components is a broader topic, so I wanted to ask if we don't do this now (in this PR), when we might expect a decision to be made on that topic?

davidsbatista: I know that nesting components is a big topic, but we could use this PR to first experiment with nested components in the context of @sjrl specific use case. Given that this is the experimental package, it's the best place to try it.

tstadel: Another option than sticking to hard types would be introducing a [protocol](https://typing.readthedocs.io/en/latest/spec/protocol.html). That way you don't need to rely on any types of the main package as long as you can specify the interface.

davidsbatista: @tstadel that's a good point and we discussed about the need of Abstractions for Generators, it would make cases like this one easier to deal with. It would be beneficial to have something similar like we have for DocumentStores. 

The extra work involves guaranteeing that every LLM underneath works with the defined interface and that you can easily switch LLMs without anything breaking, but I think is the right direction.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,257 @@
...
+        Extract metadata from documents using a Language Model.
+
+        :param documents: List of documents to extract metadata from.
+        :returns:
+            A dictionary with the key "documents_meta" containing the documents with extracted metadata.
+        """
+        errors = []
+        for document in documents:
+            prompt_with_doc = self.builder.run(input_text=document.content)
+            llm_provider = self._init_generator(self.generator_api, self.generator_api_params)
```
sjrl, line 246: Is there a reason we are reinitializing the generator in the for loop? I would think it would be better to initialize it once the init method or in a warm_up method. What do you think?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,257 @@
...
+            Dictionary with serialized data.
+        """
+        return default_to_dict(
+            self,
+            prompt=self.prompt,
+            input_text=self.input_text,
+            expected_keys=self.expected_keys,
+            raise_on_failure=self.raise_on_failure,
+            generator_api=self.generator_api.value,
+            generator_api_params=self.generator_api_params,
```
sjrl, line 219: Out of curiosity don't we need to handle the api key specifically when converting `to_dict`? In our generators we typically would call 
```python
api_key=self.api_key.to_dict()
```

Perhaps instead of doing this we could get all api_params by doing something like
```
llm_dict = llm_provider.to_dict()
generator_api_params = llm_dict["init_params"]
```
this way we could leverage the `to_dict` and `from_dict` methods of the underlying generator which handles the sede of the API key secret right?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,257 @@
...
+    def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
+        """
+        Deserializes the component from a dictionary.
+
+        :param data:
+            Dictionary with serialized data.
+        :returns:
+            An instance of the component.
+        """
+        return default_from_dict(cls, data)
```
sjrl, line 232: Similar comment here to as above, it seems like utilizing the `from_dict` method of the underlying LLM Provider would allow us to properly handle the deserialization of the API Key secret.

pyproject.toml
```diff
@@ -59,6 +59,9 @@ extra-dependencies = [
   "fastapi",
   # Tool
   "jsonschema",
+  # LLMMetadataExtractor dependencies
+  "amazon-bedrock-haystack==1.0.2",
+  "google-vertex-haystack==2.0.0",
```
sjrl, line 64: Maybe we should make these `>=` to be consistent with how we are handling the other deps here? It seems like aren't fixing the versions of other haystack integration libraries.

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: no reason at all, just a mistake from my side, thanks for point it out!

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: You are right, we are not handling the key, assuming the user will use the default one.

The question is whether we want to serialise an LLM provider that has already been initialised or simply the `generator_api` of the LLM provider and the `generator_api_params`, and then the `from_dict()` will use this information to initialise it when loaded.

Your approach would tie the `LLMMetadataExtractor` to an LLM provider when serialised/deserialised.

You are the main user @sjrl :) Let me know how you prefer it.

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: see comment above

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,277 @@
...
+        self.builder = PromptBuilder(prompt, required_variables=[input_text])
+        self.raise_on_failure = raise_on_failure
+        self.expected_keys = expected_keys
+        self.generator_api = generator_api
+        self.generator_api_params = generator_api_params or {}
+        self.llm_provider = self._init_generator(generator_api, self.generator_api_params)
+
+    def _check_prompt(self):
+        if self.input_text not in self.prompt:
+            raise ValueError(f"{self.input_text} must be in the prompt.")
```
sjrl, line 163: @davidsbatista I noticed this function isn't being called anywhere. I think it would make sense to add to the init method right?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,277 @@
...
+
+        :param data:
+            Dictionary with serialized data.
+        :returns:
+            An instance of the component.
+        """
+
+        data["init_parameters"]["generator_api"] = LLMProvider.from_str(data["init_parameters"]["generator_api"])
+        deserialize_secrets_inplace(data["init_parameters"]["generator_api_params"], keys=["api_key"])
+        return default_from_dict(cls, data)
```
sjrl, line 252: Hey, let's modify this slightly so we don't accientally require `generator_api_params` to be specified in the YAML even though it is an optional parameter. 
```suggestion
        init_parameters = data.get("init_parameters", {})
        if "generator_api" in init_parameters:
            data["init_parameters"]["generator_api"] = LLMProvider.from_str(data["init_parameters"]["generator_api"])
        if "generator_api_params" in init_parameters:
            deserialize_secrets_inplace(data["init_parameters"]["generator_api_params"], keys=["api_key"])
        return default_from_dict(cls, data)
```

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: yes, this was forgotten somehow after the changes in the init

sjrl (approved):
Thanks for the work on this!

shadeMe (changes requested):

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -257,13 +241,15 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
 
 
     @component.output_types(documents=List[Document], errors=List[Tuple[str,Any]])
```
shadeMe, line 243:
```suggestion
    @component.output_types(documents=List[Document], errors=List[Optional[Tuple[str,Any]]])
```
That said, why do we need a document id at all in `errors` - looks `documents` and `errors` are parallel arrays, so each index should correspond.

shadeMe (changes requested):
Looks like you've missed a comment above.

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: good point, I change it to remove the `id`

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -0,0 +1,281 @@
...
+    >>
+    ```
+    """ # noqa: E501
+
+    def __init__( # pylint: disable=R0917
+        self,
+        prompt: str,
+        input_text: str,
+        expected_keys: List[str],
+        generator_api: LLMProvider,
```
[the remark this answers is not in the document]
davidsbatista: I don't think this is correct, the `generator_api` only accepts a `LLMProvider`, why a string?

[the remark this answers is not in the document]
shadeMe: We should use this pattern - https://github.com/deepset-ai/haystack-core-integrations/blob/c0750ec01c8689e0d86e193223f3a2d807a3a32b/integrations/optimum/src/haystack_integrations/components/embedders/optimum/optimum_text_embedder.py#L42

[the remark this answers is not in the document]
davidsbatista: I don't understand how this pattern relates to this use case, this will affect the serialization; what is supposed to happen if a user passes a str instead of an Enum?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -140,15 +140,16 @@ def __init__( # pylint: disable=R0917
...
+            else LLMProvider.from_str(generator_api)
         self.generator_api_params = generator_api_params or {}
         self.llm_provider = self._init_generator(generator_api, self.generator_api_params)
         if self.input_text not in self.prompt:
             raise ValueError(f"Input text '{self.input_text}' must be in the prompt.")
 
     @staticmethod
     def _init_generator(
-            generator_api: LLMProvider,
+            generator_api: Union[str,LLMProvider],
```
shadeMe, line 152:
```suggestion
            generator_api: LLMProvider,
```

shadeMe (approved):

PR #102, opened 2024-09-25
ci: separate jobs for unit and integration tests

### LINKED ISSUE #86: CI test work flow should separate unit and integration steps
Currently, the CI runs both unit and integration tests under the `Unit Tests` steps. We should separate this like it's done in the core repo.

anakin87:
### Related Issues

- fixes #86

### Proposed Changes:
- run only unit tests in the existing `unit-tests` job
- create a similar job for integration tests

### How did you test it?
CI

### Notes for the reviewer
With this PR, unit tests and integration tests can run in parallel and depend on the `linting` job.
We can make different choices, but this seemed reasonable to me to avoid long running times.

I also experimented with cache (with an approach similar to that used in Haystack). Still, I encountered some issues and I also found out that restoring the cache is slow in some environments (Windows), so I decided to avoid it for the time being.

### Checklist

- I documented my code


haystack_experimental/components/tools/openapi/_openapi.py
```diff
@@ -131,7 +131,7 @@ def __init__(
         request_sender: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
         llm_provider: LLMProvider = LLMProvider.OPENAI,
         operations_filter: Optional[Callable[[Dict[str, Any]], bool]] = None,
-    ):  # noqa: PLR0913 # pylint: disable=too-many-positional-arguments
+    ):  # noqa: PLR0913
```
anakin87, line 134: This and the following change revert #101.

They seem unrelated to this PR, but probably stem from the fact that we were not using `actions/setup-python@v5` in the `linting` job, so Python (and `pylint`) versions may have been different than expected. The current version of `pylint` does not recognize this argument: `Unknown option value for '--disable', expected a valid pylint message and got 'too-many-positional-arguments'`.

shadeMe (approved):

PR #105, opened 2024-09-26
chore: Update OpenAPITool tests

vblagoje:
- adds skipif marker to `test_allowed_operations` unit test (fork repo sensitive test)
- mark `test_run_live_meteo_forecast_with_non_normalized_operation_id` as `unstable` as meteo service can be often down


vblagoje:
> Just an observation: while working on #102, I noticed that `unstable` tests are never executed and kept this behavior.

Yes, the idea (Madeesh and I talked about it) was - don't execute unstable tests on CI, but execute them locally. I'll revisit these tests as soon as I'm done with this Hackathon prep.

shadeMe (approved):

PR #106, opened 2024-09-26
feat: support for tools in `OllamaChatGenerator`

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/8190

### Proposed Changes:

I'm adding support for tools in `OllamaChatGenerator`. Similar to OpenAI PR #57.

I'm trying to respect a balance between introducing new things and being compatible with existing code.

### How did you test it?
CI, new tests.

### Notes for the reviewer
I manually tested the Chat Generator using our [notebook](https://github.com/deepset-ai/haystack-experimental/blob/main/examples/tools_support.ipynb).
The implementation generally works well.

The simple use case and processing tool results with the Chat Generator are fully reproducible.

Reproducing the multi-tool chat application (with human in the loop) is not always stable with `llama3.2:3b`/`llama3.1:8b`.
The model tends to produce tool calls (even if not needed) and sometimes hallucinates tools.
Also, Ollama support for tools is not fully polished yet. (https://github.com/ollama/ollama/issues/6127)

### Checklist

- I documented my code

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ env:
   COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
   FIRECRAWL_API_KEY: ${{ secrets.FIRECRAWL_API_KEY }}
   SERPERDEV_API_KEY: ${{ secrets.SERPERDEV_API_KEY }}
-  OLLAMA_LLM_FOR_TESTS: "qwen2.5:3b"
+  OLLAMA_LLM_FOR_TESTS: "llama3.2:3b"
 
 jobs:
   linting:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ env:
   COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
   FIRECRAWL_API_KEY: ${{ secrets.FIRECRAWL_API_KEY }}
   SERPERDEV_API_KEY: ${{ secrets.SERPERDEV_API_KEY }}
-  OLLAMA_LLM_FOR_TESTS: "llama3.2:1b"
+  OLLAMA_LLM_FOR_TESTS: "qwen2.5:3b"
 
 jobs:
   linting:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,7 @@ env:
   COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
   FIRECRAWL_API_KEY: ${{ secrets.FIRECRAWL_API_KEY }}
   SERPERDEV_API_KEY: ${{ secrets.SERPERDEV_API_KEY }}
+  OLLAMA_LLM_FOR_TESTS: "llama3.2:1b"
 
 jobs:
   linting:
@@ -116,6 +117,27 @@ jobs:
 
       - name: Install Hatch
         run: pip install hatch==${{ env.HATCH_VERSION }}
+
+      - name: Install Ollama and pull the required models
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          curl -fsSL https://ollama.com/install.sh | sh
+          ollama serve &
+
+          # Check if the service is up and running with a timeout of 60 seconds
+          timeout=60
+          while [ $timeout -gt 0 ] && ! curl -sSf http://localhost:11434/ > /dev/null; do
+            echo "Waiting for Ollama service to start..."
+            sleep 5
+            ((timeout-=5))
+          done
+
+          if [ $timeout -eq 0 ]; then
+            echo "Timed out waiting for Ollama service to start."
+            exit 1
+          fi
+
+          ollama pull ${{ env.OLLAMA_LLM_FOR_TESTS }}
 
       - name: Run
         run: hatch run test:integration
--- a/haystack_experimental/components/generators/ollama/chat/chat_generator.py
+++ b/haystack_experimental/components/generators/ollama/chat/chat_generator.py
@@ -1,0 +1,183 @@
+# from haystack.dataclasses import ChatMessage, StreamingChunk
+from typing import Any, Callable, Dict, List, Optional, Type, Union
+
+from haystack import component, default_from_dict
+from haystack.dataclasses import StreamingChunk
+from haystack.lazy_imports import LazyImport
+from haystack.utils.callable_serialization import deserialize_callable, serialize_callable
+
+from haystack_experimental.dataclasses import ChatMessage, ToolCall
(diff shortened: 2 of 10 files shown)


haystack_experimental/components/generators/ollama/chat/chat_generator.py
```diff
@@ -0,0 +1,244 @@
...
+
+with LazyImport("Run 'pip install ollama-haystack'") as ollama_integration_import:
+    # pylint: disable=import-error
+    from haystack_integrations.components.generators.ollama import OllamaChatGenerator as OllamaChatGeneratorBase
+
+
+if ollama_integration_import.is_successful():
+    chatgenerator_base_class: Type[OllamaChatGeneratorBase] = OllamaChatGeneratorBase
+else:
+    chatgenerator_base_class: Type[object] = object  # type: ignore[no-redef]
```
anakin87, line 23: This ensures that people can use `haystack-experimental` without installing `ollama-haystack`.
`# type: ignore[no-redef]` is not nice but with different approaches, mypy raises many more errors.

(discussed a bit with Silvano)

.github/workflows/tests.yml
```diff
@@ -35,6 +35,7 @@ env:
   COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
   FIRECRAWL_API_KEY: ${{ secrets.FIRECRAWL_API_KEY }}
   SERPERDEV_API_KEY: ${{ secrets.SERPERDEV_API_KEY }}
+  OLLAMA_LLM_FOR_TESTS: "llama3.2:3b"
```
anakin87, line 38: small model that properly supports tool calling

haystack_experimental/components/generators/ollama/chat/chat_generator.py
```diff
@@ -0,0 +1,244 @@
...
+            The number of seconds before throwing a timeout error from the Ollama API.
+        :param streaming_callback:
+            A callback function that is called when a new token is received from the stream.
+            The callback function accepts StreamingChunk as an argument.
+        :param tools:
+            A list of tools for which the model can prepare calls.
+            Not all models support tools. For a list of models compatible with tools, see the
+            [models page](https://ollama.com/search?c=tools).
+        """
+        ollama_integration_import.check()
```
vblagoje, line 119: Isn't this going to fail if we are extending object instead of OllamaChatGenerator

haystack_experimental/components/generators/ollama/chat/chat_generator.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        Deserialize this component from a dictionary.
+
+        :param data: The dictionary representation of this component.
+        :returns:
+            The deserialized component instance.
+        """
+        deserialize_tools_inplace(data["init_parameters"], key="tools")
+        init_params = data.get("init_parameters", {})
+        serialized_callback_handler = init_params.get("streaming_callback")
```
vblagoje, line 158: Something seems off here. Isn't super class taking care of these callback handlers?

.github/workflows/tests.yml
```diff
@@ -117,5 +118,26 @@ jobs:
       - name: Install Hatch
         run: pip install hatch==${{ env.HATCH_VERSION }}
 
+      - name: Install Ollama and pull the required models
```
vblagoje, line 121: Does it make sense to have this step run only on integration tests?

vblagoje (commented):
@anakin87 I did only a rough pass first to provide some feedback, hopefully useful :-) I'll do another after this one.

haystack_experimental/components/generators/ollama/chat/chat_generator.py
anakin87: We discussed this for the `OpenAIChatGenerator` and decided to keep the full method.
https://github.com/deepset-ai/haystack-experimental/pull/57#discussion_r1718266866

.github/workflows/tests.yml
anakin87: This step takes some time (about 1 minute), so I think it makes sense to skip it for unit tests that do not require an Ollama instance running.

haystack_experimental/components/generators/ollama/chat/chat_generator.py
vblagoje: Ok gotcha now, I was trying to figure out why we extend object when `not ollama_integration_import.is_successful()` and we basically do that only to enable smooth functioning of this `ollama-haystack` installed or not import mechanism, correct? If so, perhaps one line of comment would help with maintenance here so we don't forget. Perhaps only commenting with the above [comment](https://github.com/deepset-ai/haystack-experimental/pull/106#discussion_r1782772497).

haystack_experimental/components/generators/ollama/chat/chat_generator.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def _build_message_from_ollama_response(self, ollama_response: Dict[str, Any]) -> ChatMessage:
+        """
+        Converts the non-streaming response from the Ollama API to a ChatMessage.
+        """
+        ollama_message = ollama_response["message"]
+
+        text = ollama_message["content"]
+
+        tool_calls = []
```
vblagoje, line 172: How about?
```
tool_calls = [
            ToolCall(tool_name=tc["function"]["name"], arguments=tc["function"]["arguments"])
            for tc in ollama_message.get("tool_calls", [])
        ]
```

Should be the same but simpler?

haystack_experimental/components/generators/ollama/chat/chat_generator.py
```diff
@@ -0,0 +1,244 @@
...
+
+    if not text_contents and not tool_calls and not tool_call_results:
+        raise ValueError("A `ChatMessage` must contain at least one `TextContent`, `ToolCall`, or `ToolCallResult`.")
+    elif len(text_contents) + len(tool_call_results) > 1:
+        raise ValueError("A `ChatMessage` can only contain one `TextContent` or one `ToolCallResult`.")
+
+    ollama_msg: Dict[str, Any] = {"role": message._role.value}
+
+    if tool_call_results:
+        result = tool_call_results[0]
```
vblagoje, line 42: a bit of nit, why even use this result variable at all?

vblagoje (approved):
Seems fine to me, small suggestions here and there.

haystack_experimental/components/generators/ollama/chat/chat_generator.py
anakin87: I will add a comment explaining this.

haystack_experimental/components/generators/ollama/chat/chat_generator.py
anakin87: Right! I will do something similar...

shadeMe (approved):

PR #118, opened 2024-10-16
feat: support for tools in `AnthropicChatGenerator`

vblagoje:
Adds AnthropicChatGenerator with tools support

fixes https://github.com/deepset-ai/haystack/issues/8261


haystack_experimental/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -0,0 +1,393 @@
...
+        # update generation kwargs by merging with the generation kwargs passed to the run method
+        generation_kwargs = {**self.generation_kwargs, **(generation_kwargs or {})}
+        disallowed_params = set(generation_kwargs) - set(self.ALLOWED_PARAMS)
+        if disallowed_params:
+            logger.warning(
+                f"Model parameters {disallowed_params} are not allowed and will be ignored. "
+                f"Allowed parameters are {self.ALLOWED_PARAMS}."
+            )
+        generation_kwargs = {k: v for k, v in generation_kwargs.items() if k in self.ALLOWED_PARAMS}
+        tools = tools or self.tools
```
anakin87, line 340: let's check duplicate tools
something like: https://github.com/deepset-ai/haystack-experimental/blob/52d7567524de595ed4b6ede275a03b5c4d7a09da/haystack_experimental/components/generators/chat/openai.py#L257-L261

anakin87 (changes requested):
I found some opportunities for improvement and clarification.

Other than that:
- I think some unit tests can be added, to cover the introduced utility functions.
- let's export the component to haystack_experimental/components/__init__.py
- update the pydoc config in docs/pydoc/config/generators_api.yml

haystack_experimental/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -0,0 +1,397 @@
...
+            nonlocal current_tool_call
+            # this is the tool call start, capture the id and name
+            # arguments will be captured in the delta
+            if chunk.meta.get("content_block", {}).get("type") == "tool_use":
+                delta_block = chunk.meta.get("content_block")
+                current_tool_call = {
+                    "id": delta_block.get("id"),
+                    "name": delta_block.get("name"),
+                    "arguments": "",
+                }
```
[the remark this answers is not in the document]
vblagoje: Just for readability, but I'll simplify and won't used nested functions

haystack_experimental/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -0,0 +1,397 @@
...
+                tool_calls.append(
+                    ToolCall(
+                        tool_name=content_block.name,
+                        arguments=content_block.input,  # dict already
+                        id=content_block.id,
+                    )
+                )
+
+        # Combine all text parts into a single string
+        full_text = " ".join(text_parts)
```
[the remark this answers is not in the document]
vblagoje: I could swear somewhere that I've seen it in docs - but you're right. Will remove this

vblagoje:
@anakin87 Antropic is a bit particular regarding parallel tool invocation as it requires tool_result of parallel calls to be packed in a single message. Not sure if that is the case with other generators but it is here. When I was sending one response message per tool - LLM wouldn't accept the message list complaining that it couldn't pair the response to tool id. Once I started packing both tool responses into a single message - it [worked](https://github.com/deepset-ai/haystack-experimental/pull/118/commits/69cd91e56e56984657d9451bb3882d10f5151c46). The error for each tool response per message was:
```
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': "messages.2: the following `tool_use` ids were not found in `tool_result` blocks: {'toolu_019sLMuA1TfFJSz6Vofmg3Hn'}"}}```

**Comment by anakin87** [comment]
> @anakin87 Antropic is a bit particular regarding parallel tool invocation as it requires tool_result of parallel calls to be packed in a single message. Not sure if that is the case with other generators but it is here. When I was sending one response message per tool - LLM wouldn't accept the message list complaining that it couldn't pair the response to tool id. Once I started packing both tool responses into a single message - it [worked](https://github.com/deepset-ai/haystack-experimental/pull/118/commits/69cd91e56e56984657d9451bb3882d10f5151c46). The error for each tool response per message was:
> 
> ```
> anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': "messages.2: the following `tool_use` ids were not found in `tool_result` blocks: {'toolu_019sLMuA1TfFJSz6Vofmg3Hn'}"}}```
> ```

Ah, I understand. As far as I know, this does not happen for other providers I worked with...

**Comment by anakin87** [comment]
With respect to https://github.com/deepset-ai/haystack-experimental/pull/118/commits/69cd91e56e56984657d9451bb3882d10f5151c46, do you think that something like this would work?

- replace ` _convert_message_to_anthropic_format` with `_convert_messages_to_anthropic_format(message: List[ChatMessage]) -> List[Dict[str, Any]]`
- handle packing of tool results in the the new function (`_convert_messages_to_anthropic_format`)

**Comment by vblagoje** [comment]
> With respect to [69cd91e](https://github.com/deepset-ai/haystack-experimental/commit/69cd91e56e56984657d9451bb3882d10f5151c46), do you think that something like this would work?
> 
> * replace ` _convert_message_to_anthropic_format` with `_convert_messages_to_anthropic_format(message: List[ChatMessage]) -> List[Dict[str, Any]]`
> * handle packing of tool results in the the new function (`_convert_messages_to_anthropic_format`)

In that approach we can reason about the message order and state overall and "intelligently" convert a list of messages as we see fit. But what do we gain?

**Comment by anakin87** [comment]
I think we would gain compatibility across Generators and with the `ToolInvoker`.
For example, OpenAI and Ollama do not support messages with more than one tool result (see our [OpenAI code](https://github.com/deepset-ai/haystack-experimental/blob/77ae05622b2d6368f63b925b87e715733e6f910e/haystack_experimental/components/generators/chat/openai.py#L33)).

**Comment by vblagoje** [comment]
@anakin87 I've done some basic work without polishing the details to get going in the agreed direction. Any idea what's up with CI?

**Comment by anakin87** [comment]
I created a PR to fix the issue: https://github.com/deepset-ai/haystack-experimental/pull/126
Let's see if it works.

**Comment by vblagoje** [comment]
Looks good @anakin87 - I'll test out your notebook with Anthropic and do another pass to polish this latest stuff. Should be gtg tomorrow

**Comment by vblagoje** [comment]
Here is the test [notebook](https://colab.research.google.com/drive/1d0Ow-Sy0l6Koz5Hf3TUvC9lneg74iJ1L?usp=sharing) that implements the original tool_support [notebook](https://github.com/deepset-ai/haystack-experimental/blob/main/examples/tools_support.ipynb) using `AnthorpicChatGenerator`.

File: haystack_experimental/components/generators/anthropic/chat/chat_generator.py [hunk 4]
```diff
@@ -0,0 +1,468 @@
...
+        :param generation_kwargs: Additional parameters to use for the model. These parameters are sent directly to
+            the Anthropic API. See Anthropic's documentation for more details on available parameters.
+             Supported generation_kwargs parameters are:
+            - `system`: The system message to be passed to the model.
+            - `max_tokens`: The maximum number of tokens to generate.
+            - `metadata`: A dictionary of metadata to be passed to the model.
+            - `stop_sequences`: A list of strings that the model should stop generating at.
+            - `temperature`: The temperature to use for sampling.
+            - `top_p`: The top_p value to use for nucleus sampling.
+            - `top_k`: The top_k value to use for top-k sampling.
```
anakin87, line 233: This part is not copied from https://github.com/deepset-ai/haystack-core-integrations/blob/7eb062c63dcb6dfa6bbaa4ba393e76ff39715ac4/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py#L89
and there are some errors.

anakin87 (changes requested):
This PR is almost ready to be merged.
I found only a few possible simplifications...

pyproject.toml
```diff
@@ -77,7 +77,7 @@ lint = [
   "pylint -ry -j 0 {args:haystack_experimental}",
 ]
 test-cov = "coverage run -m pytest -m \"not unstable\" {args:test}"
-cov-report = ["- coverage combine", "coverage xml"]
+cov-report = ["- coverage combine", "coverage xml", "coverage html"]
```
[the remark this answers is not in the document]
vblagoje: oops it is not, it snuck in when I was working on coverare 👍

haystack_experimental/components/generators/anthropic/chat/chat_generator.py
vblagoje: Yeah, I copied most but not extra_headers. As that's related to prompt cache-ing. Do you think we should have 1:1 match between the two versions here?

In which sense errors?

anakin87:
```
        :param api_key: The Anthropic API key. You can set it with an environment variable `ANTHROPIC_API_KEY`,
            or pass it directly during initialization.
        :param model: The name of the Anthropic model to use. Defaults to "claude-3-opus-20240229".
```

- API key cannot be passed directly (it must be wrapped in a `Secret`).
- We usually do not include defaults in docstring. In any case, the default model referenced in the docstring is not the same as in the code.

anakin87:
Please take into account [this comment](https://github.com/deepset-ai/haystack-experimental/pull/118#discussion_r1819224356)

anakin87 (approved):
Messages conversion: I further simplified the logic and added a test.

I think it's good to go now!

PR #125, opened 2024-10-24
feat: LLMMetadata Extractor

davidsbatista:
### Related Issues

- fixes [#8450](https://github.com/deepset-ai/haystack/issues/8450)

### Proposed Changes:

- adds a new component to extract metadata from a doc and update the doc based on a prompt

### How did you test it?

- unit tests, integration tests, manual verification

### Checklist

- I documented my code


anakin87:
I created a PR to fix the failing unit tests: https://github.com/deepset-ai/haystack-experimental/pull/126

sjrl:
I know not directly related to this PR, but I wonder for these lines in the init method

https://github.com/deepset-ai/haystack-experimental/blob/a099dcda93c2be27936f5e418869b1421325536a/haystack_experimental/components/extractors/llm_metadata_extractor.py#L121-L122

if `input_text` should be renamed to `prompt_variable`? Just when coming back to this component right now I realized I was a bit confused by the naming and doc string for `input_text` and it only clicked when I read through your example. What do you think?

haystack_experimental/util/utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # add remaining keys from d2
+    for key in d2.keys():
+        if key not in merged:
+            merged[key] = d2[key]
+
+    return merged
+
+
+def expand_page_range(page_range: Union[ List[Union[int,str]], List[int] ] ):
```
sjrl, line 49: Thanks for adding this! One thing that would be nice is if we also support an example like 
```
['1-3', '5', '8', '10-12']
```
where the whole input is a list of strings. Currently when trying this example I get an error at this line
```python
start, end = page.split("-")
```

sjrl:
Also rereading the run function I noticed https://github.com/deepset-ai/haystack-experimental/blob/a099dcda93c2be27936f5e418869b1421325536a/haystack_experimental/components/extractors/llm_metadata_extractor.py#L275

which I think is a bug, since we are assuming the name of the prompt variable is `input_text` instead of using the name provided by the user in the init method right?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +242,54 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
             result = self.llm_provider.run(prompt=prompt_with_doc["prompt"])
             llm_answer = result["replies"][0]
             if self.is_valid_json_and_has_expected_keys(expected=self.expected_keys, received=llm_answer):
                 extracted_metadata = json.loads(llm_answer)
                 for k in self.expected_keys:
                     document.meta[k] = extracted_metadata[k]
+                # if in extra_from_range mode, merge all extracted in a single dict
+                if extract_from_range and extracted_metadata:
+                    all_metadata = merge_dicts(all_metadata, extracted_metadata)
+                    print(all_metadata)
```
sjrl, line 285: I think this is a left over print

haystack_experimental/util/utils.py
davidsbatista: so you are suggesting, essentially, everything as a string

sjrl: Yeah support:
* everything as a string
* everything as a number (when no ranges are specified)
* and I guess the mixture of both

davidsbatista:
@sjrl I've fixed the issue with the prompt builder/variables, and refactored substantially the code to accommodate for the use case you specified

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -127,25 +130,27 @@ def __init__( # pylint: disable=R0917
...
         self.raise_on_failure = raise_on_failure
         self.expected_keys = expected_keys
         self.generator_api = generator_api if isinstance(generator_api, LLMProvider)\
             else LLMProvider.from_str(generator_api)
         self.generator_api_params = generator_api_params or {}
         self.llm_provider = self._init_generator(self.generator_api, self.generator_api_params)
-        if self.input_text not in self.prompt:
-            raise ValueError(f"Input text '{self.input_text}' must be in the prompt.")
+        if self.prompt_variable not in self.prompt:
+            raise ValueError(f"Input text '{self.prompt_variable}' must be in the prompt.")
```
sjrl, line 152:
```suggestion
            raise ValueError(f"Prompt variable '{self.prompt_variable}' must be in the prompt.")
```

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +245,73 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+        Extract metadata from the content and update the document with the extracted metadata. If the extraction fails,
+        the error message will be stored in `llm_errors`.
+
+        :param document: Document to be updated with the extracted metadata.
+        :param errors: Dictionary to store error messages if the extraction fails.
+        :param content: Content to extract metadata from.
+        """
+        prompt_with_doc = self.builder.run(
+            template=self.prompt,
+            template_variables={self.prompt_variable:content}
```
sjrl, line 259: I'm a little confused by this line, do we need to set `template_variables`?

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +245,73 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+        )
+        print(f"Prompt with doc: {prompt_with_doc}")
```
sjrl, line 261: Let's remove the print statement or if we want to keep let's make it a logging statement.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +245,73 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+            extracted_metadata = json.loads(llm_answer)
+            for k in self.expected_keys:
+                document.meta[k] = extracted_metadata[k]
+        else:
+            errors[document.id] = llm_answer
+
 
     @component.output_types(documents=List[Document], errors=Dict[str, Any])
-    def run(self, documents: List[Document]) -> Dict[str, Any]:
+    def run(self, documents: List[Document], page_range: Optional[List[str]] = None):    # pylint: disable=R0914
```
sjrl, line 273: Could we also add `page_range` as an init parameter? For me I'd probably only use it there and not so much in the run method. So I think it would be good to have in both.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +245,73 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+                logger.warning(f"Document {document.id} has no content. Skipping metadata extraction.")
+                continue
+
+            if page_range:  # extract metadata from a specific range of pages
+                expanded_page_range = expand_page_range(page_range) if page_range else None
+                start_document = expanded_page_range[0] - 1
+                end_document = expanded_page_range[-1]
+                splitter = DocumentSplitter(split_by="page", split_length=1)
+                pages = splitter.run(documents=[document])
+                merged_contents = [page.content + '\n' for page in pages['documents'][start_document:end_document]]
```
sjrl, line 310: This doesn't respect that the page ranges specified could be non-continuous. For example if I specify `page_range=['2', '5']` this code will send pages 2, 3, 4 and 5 and not just 2 and 5.

haystack_experimental/util/utils.py
```diff
@@ -0,0 +1,81 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Union, List
+
+
+def merge_dicts(d1, d2):
```
sjrl, line 8: I don't think this function is used anymore, so we could remove right?

haystack_experimental/components/extractors/llm_metadata_extractor.py
sjrl: Ahh okay I gotcha we are mapping `content` to `self.prompt_variable`. Apologies, missed that point.

davidsbatista: no need to apologize, it's good that you are thorough on the reviews 👍🏽

davidsbatista:
@sjrl let me know if need anything else, otherwise I would now focus on writing comprehensive tests

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +253,80 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+            template_variables={self.prompt_variable:content}
+        )
+        result = self.llm_provider.run(prompt=prompt_with_doc["prompt"])
+        llm_answer = result["replies"][0]
+        if self.is_valid_json_and_has_expected_keys(expected=self.expected_keys, received=llm_answer):
+            extracted_metadata = json.loads(llm_answer)
+            for k in self.expected_keys:
+                document.meta[k] = extracted_metadata[k]
+        else:
+            errors[document.id] = llm_answer
```
sjrl, line 278: I don't think this is needed now, but in the future also adding in the error thrown by the JSON loader would be helpful to understand why the `llm_answer` isn't working.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +253,80 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+
+            if not document.content:
+                logger.warning(f"Document {document.id} has no content. Skipping metadata extraction.")
+                continue
+
+            if self.expanded_range or page_range: # extract metadata from a specific range of pages
+
+                # override the page range if provided
+                if page_range:
+                    self.expanded_range = expand_page_range(page_range)
```
sjrl, line 316: I would rewrite this section to follow how we usually handle overwriting runtime parameters with init parameters. Basically we should never set the `self.expanded_range` in the run method, it should always be left to whatever it was initialized to. So something like 
```python
expanded_range = self.expanded_range
if page_range:
    expanded_range = expand_page_range(page_range)
...
```

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +253,80 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+
+                if not self.expanded_range:
+                    msg = f"Page range {self.expanded_range} invalid"
+                    raise ValueError(msg)
```
sjrl, line 320: I think this is an error that should be raised in the `expand_page_range` function. We should already know in that function if the provided `page_range` was invalid or not.

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -240,28 +253,80 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+
+                splitter = DocumentSplitter(split_by="page", split_length=1)
+                pages = splitter.run(documents=[document])
+                content = [p.content + "\n" for idx, p in enumerate(pages["documents"]) if idx in self.expanded_range]
```
sjrl, line 324: We should reinsert the page break and not a newline and the idx starts at 0 whereas the `expanded_range` seems to start at 1. So this would send the wrong pages, shifted all by 1. So change to 
```suggestion
                content = [p.content + "\f" for idx, p in enumerate(pages["documents"]) if (idx + 1) in self.expanded_range]
```

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: makes sense 👍🏽

haystack_experimental/components/extractors/llm_metadata_extractor.py
davidsbatista: I'm aware of that, I left this check here because of the linter and the type checker - it seems they are not robust enough to understand check that in run time if `expanded_range` == [] it will never reach the line with the for loop to extract the pages content - let me see if I can fix it in some other way

sjrl: ahh okay, if it's too much hastle then fine to leave in, thanks for the explanation!

sjrl (approved):
Looks good!

PR #147, opened 2024-12-05
feat: Metadata extractor update

sjrl:
### Related Issues

Updates to the LLM Metadata Extractor. 

### Proposed Changes:

 
 
**Output Changes:**
- Returns `documents` which contains all documents that succeeded in metadata extraction and `failed_documents` which contains all documents that failed (either because the LLM could not be executed or a JSON parsing error). By having this separation it's easy to use the `failed_documents` directly in another LLM metadata extractor to try and fix the issue especially if the error is related to JSON parsing. 

**Prompt Changes**
- Removes the prompt_variable 
- Instead we require the prompt variable to have to be `document`. This way we can utilize all the info contained within a document as well. E.g. I can use `document.content` to print out the contents and then also use `document.meta.XX` to print out any relevant meta information as well (e.g. file name) into the prompt instructions. 

**Se/de**
- Identified some bugs in the from_dict method. Basically, we don't have a way where we can automatically call the `from_dict` method of all the different LLM Providers. So instead we need to create a custom from_dict method that can optionally handle all the different edge cases. 

**Other**
- Made `expected_keys` optional since it's not always alarming if not all expected keys are filled. Sometimes documents won't contain relevant info to extract so instead it can be passed optionally. If passed now it will only print a warning message if there are missing (or additionally unexpected keys). 
- Used Thread Pool Executor to add "parallelism" to the LLM calls. This should greatly speed up the component since it will run the LLM calls in parallel. 

### How did you test it?

Added a lot of new tests

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/extractors/llm_metadata_extractor.py
+++ b/haystack_experimental/components/extractors/llm_metadata_extractor.py
@@ -70,7 +70,7 @@ class LLMMetadataExtractor:
     The component will run the LLM on each document in the list and extract metadata from the document. The metadata
     will be added to the document's metadata field. If the LLM fails to extract metadata from a document, the document
     will be added to the `failed_documents` list. The failed documents will have the keys `metadata_extraction_error` and
-    `metadata_extraction_response` in their metadata. These documents can be re-run with the another extractor to
+    `metadata_extraction_response` in their metadata. These documents can be re-run with another extractor to
     extract metadata by using the `metadata_extraction_response` and `metadata_extraction_error` in the prompt.
 
     ```python
--- a/haystack_experimental/components/extractors/llm_metadata_extractor.py
+++ b/haystack_experimental/components/extractors/llm_metadata_extractor.py
@@ -335,7 +335,7 @@ class LLMMetadataExtractor:
                 content = ""
                 for idx, page in enumerate(pages["documents"]):
                     if idx + 1 in expanded_range:
-                        content += page.content + "\f"
+                        content += page.content
                 doc_copy.content = content
             else:
                 doc_copy = document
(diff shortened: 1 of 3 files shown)


haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -248,37 +264,111 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
         if "generator_api_params" in init_parameters:
-            deserialize_secrets_inplace(data["init_parameters"]["generator_api_params"], keys=["api_key"])
+            # Check all the keys that need to be deserialized
+            azure_openai_keys = ["azure_ad_token"]
+            aws_bedrock_keys = [
+                "aws_access_key_id", "aws_secret_access_key", "aws_session_token", "aws_region_name", "aws_profile_name"
+            ]
+            deserialize_secrets_inplace(
+                data["init_parameters"]["generator_api_params"],
+                keys=["api_key"] + azure_openai_keys + aws_bedrock_keys,
```
davidsbatista, line 279: don't you also need to add the OpenAI key here?

davidsbatista (commented):
Looks very good - just left a few minor comments/checks

haystack_experimental/components/extractors/llm_metadata_extractor.py
```diff
@@ -248,37 +264,111 @@ def from_dict(cls, data: Dict[str, Any]) -> "LLMMetadataExtractor":
...
+                all_prompts.append(None)
+                continue
+
+            if expanded_range:
+                doc_copy = copy.deepcopy(document)
+                pages = self.splitter.run(documents=[doc_copy])
+                content = ""
+                for idx, page in enumerate(pages["documents"]):
+                    if idx + 1 in expanded_range:
+                        content += page.content
```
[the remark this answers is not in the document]
sjrl: When I checked the output from the DocumentSplitter it already includes the page separator at the end of each chunk. I believe this is b/c the DocumentSplitter doesn't remove the split by character.

haystack_experimental/components/extractors/llm_metadata_extractor.py
sjrl: The OpenAI key is called `api_key` in the init method so it's covered.

PR #150, opened 2024-12-09
feat: Update handling of tools_strict to ensure compatibility with vLLM

Emil-io:
### Proposed Changes:
Only add the "strict" dictionary to the openai_tools for generation, if tools_strict is available.

This ensures compatibility with vLLM (which adopts the OpenAI Interface). vLLM throws an error, if the "strict" dictionary is part of the tools list:

`openai.BadRequestError: Error code: 400 - {'object': 'error', 'message': "[{'type': 'extra_forbidden', 'loc': ('body', 'tools', 0, 'function', 'strict'), 'msg': 'Extra inputs are not permitted', 'input': False}`

### Checklist

- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:` and added `!` in case the PR includes breaking changes.
- I documented my code

### DIFF
--- a/haystack_experimental/components/generators/chat/openai.py
+++ b/haystack_experimental/components/generators/chat/openai.py
@@ -447,7 +447,10 @@ class OpenAIChatGenerator:
             openai_tools = [
                 {
                     "type": "function",
-                    "function": {**t.tool_spec, "strict": tools_strict},
+                    "function": {
+                        **t.tool_spec,
+                        **({"strict": tools_strict} if tools_strict else {}),
+                    },
                 }
                 for t in tools
             ]
--- a/haystack_experimental/components/generators/chat/openai.py
+++ b/haystack_experimental/components/generators/chat/openai.py
@@ -449,7 +449,7 @@ class OpenAIChatGenerator:
                     "type": "function",
                     "function": {
                         **t.tool_spec,
-                        **({"strict": tools_strict} if tools_strict else {}),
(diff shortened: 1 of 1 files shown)


haystack_experimental/components/generators/chat/openai.py
```diff
@@ -447,7 +447,10 @@ def _prepare_api_call(  # noqa: PLR0913
             openai_tools = [
                 {
                     "type": "function",
-                    "function": {**t.tool_spec, "strict": tools_strict},
+                    "function": {
+                        **t.tool_spec,
+                        **({"strict": tools_strict} if tools_strict else {}),
```
julian-risch, line 452:
```suggestion
                        **({"strict": tools_strict} if tools_strict is not None else {}),
```

julian-risch: We should allow setting `tools_strict` to `false` and only set it to `{}` if no value was provided at all.

julian-risch (changes requested):
I have a small change request but other than that the PR looks good to me. I would like to get another review from @vblagoje or @anakin87 before we merge.

anakin87 (approved):
LGTM!

vblagoje (approved):
LGTM as well

PR #143, opened 2024-11-28
chore: Remove discontinued `OpenAPITool` and `OpenAIFunctionCaller` experiments

shadeMe:
### Proposed Changes:

 
 

Remove the discontinued experiments and update the readme.

### How did you test it?

Locally

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack/blob/main/code_of_conduct.txt)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.
- I documented my code
- I ran [pre-commit hooks](https://github.com/deepset-ai/haystack/blob/main/CONTRIBUTING.md#installation) and fixed any issue

PR #151, opened 2024-12-09
feat: Tool openapi integration

vblagoje:
### Why:
Creates `Tool` instances from OpenAPI specifications, enabling the invocation of tools based on OpenAPI JSON or YAML formats.

### What:
- Introduced the `from_openapi_spec` class method in the `Tool` class, enabling tool creation from OpenAPI specifications provided as URLs, file paths, or string content.
- Added `OpenAPIKwargs` TypedDict to encapsulate configuration options for OpenAPI integration, including API credentials and allowed operations.
- Uses https://github.com/vblagoje/openapi-llm/ library that is essentially eviscerated [code](https://github.com/deepset-ai/haystack-experimental/pull/143) from the former `OpenAPITool`

### How can it be used:
- To create tools with OpenAPI specifications:
```python
from haystack_experimental.components.tools.openapi import create_tools_from_openapi_spec

tools = create_tools_from_openapi_spec(
            spec="https://bit.ly/serperdev_openapi",
            credentials="your-api-key-here"
)
# now you can use tools just like any other Haystack tool
```
- Supports various methods for importing OpenAPI specifications, making it easy to manage different sources of API definitions.
- Ensures customizable invocation of API endpoints through dynamically generated tool definitions based on the OpenAPI schema.

### How did you test it:
- Conducted unit tests that validate the `from_openapi_spec` method's functionality and verify that tool attributes align with the OpenAPI definitions.
- See the [notebook](https://colab.research.google.com/drive/1QjyARiaY64BGXuRGgJv_-WK-6C1WNSmu?usp=sharing)
- More tests soon to be added

### Notes for the reviewer:
- Pay special attention to the implementation of the `from_openapi_spec` method to ensure comprehensive handling of various input types (URLs, file paths, string content).


vblagoje:
Before I add more tests @anakin87 please have a look to see if you like the general direction of this approach. I've designed it to require minimal plug-in point to our new `Tool`ing cc @julian-risch @Amnah199

haystack_experimental/dataclasses/tool.py
```diff
@@ -4,16 +4,27 @@
...
+logger = logging.getLogger(__name__)
+
 with LazyImport(message="Run 'pip install jsonschema'") as jsonschema_import:
     from jsonschema import Draft202012Validator
     from jsonschema.exceptions import SchemaError
 
+with LazyImport(message="Run 'pip install openapi-llm'") as openapi_llm_import:
+    from openapi_llm.client.config import ClientConfig
+    from openapi_llm.client.openapi import OpenAPIClient
+    from openapi_llm.core.spec import OpenAPISpecification
```
anakin87, line 26: In general, I see that offloading most of the logic in Vladimir's library results for us in less maintenance effort (and I am happy about that).
Nevertheless, I see some risks. @julian-risch what's your take on this aspect?

haystack_experimental/dataclasses/tool.py
```diff
@@ -196,6 +207,85 @@ def get_weather(
 
         return Tool(name=function.__name__, description=tool_description, parameters=schema, function=function)
 
+    @classmethod
+    def from_openapi_spec(cls, spec: Union[str, Path], **kwargs: OpenAPIKwargs) -> List["Tool"]:
```
anakin87, line 211: Differently from `@from_function` class method, this method returns a list of Tools. 
Calling `Tool.from_openapi_spec` and getting a list of Tools seems a bit confusing  from a semantic point of view.

Maybe is it better to create a function (not a class method): something like `create_tools_from_openapi_spec`.
WDYT? Do you have better ideas?

haystack_experimental/dataclasses/tool.py
```diff
@@ -237,3 +327,41 @@ def deserialize_tools_inplace(data: Dict[str, Any], key: str = "tools"):
             deserialized_tools.append(Tool.from_dict(tool))
 
         data[key] = deserialized_tools
+
+
+def _standardize_tool_definition(llm_specific_tool_def: Dict[str, Any]) -> Optional[Dict[str, Any]]:
```
anakin87, line 332: This utility function only ensures compatibility between Haystack and the new library. I don't expect to use it elsewhere in Haystack if we do things right.
Could it be moved to the library?

anakin87 (commented):
In general, I like this lightweight approach (mostly due to the fact that we are moving the logic to an external library).

Before moving on, I'd like @julian-risch to take a look.

haystack_experimental/dataclasses/tool.py
vblagoje: Good point. I'll move it.

vblagoje:
Refactored it @anakin87 @julian-risch

haystack_experimental/dataclasses/tool.py
```diff
@@ -198,6 +199,23 @@ def get_weather(
...
+
+        :param spec: OpenAPI specification as URL, file path, or string content
+        :param operation_name: Name of the operation to create a tool for
+        :param kwargs: Additional configuration options for the OpenAPI client:
+            - credentials: API credentials (e.g., API key, auth token)
+            - request_sender: Custom callable to send HTTP requests
+        :returns: Tool instance for the specified operation
+        :raises ValueError: If the OpenAPI specification is invalid or cannot be loaded
+        """
+        from haystack_experimental.tools.openapi import _create_tool_from_openapi_spec
```
vblagoje, line 215: Circular dep between these two otherwise

vblagoje:
Here is another idea to think about if we are to follow the Tool.from pattern. This method requires openapi spec and operation name and creates only one Tool just like other from methods. Just an idea....

julian-risch (changes requested):
With the recent refactoring of Tool class and the introduction of ComponentTool, we agreed that OpenAPITool and/or an OpenAPI component is the best way forward

vblagoje:
Closing, new ideas emerged.

PR #159, opened 2024-12-18
feat: Add Tool.from_component

vblagoje:
### Why:
Automates the conversion of Haystack components into LLM tools. 

- fixes https://github.com/deepset-ai/haystack/issues/8630

### What:
- Added `component_schema.py`, that converts component run parameters into a JSON tools schema format.
- Modified `tool.py` to introduce a `from_component` method, enabling the creation of `Tool` instances from Haystack components. This makes tool creation and integration into pipelines more dynamic.
- Added comprehensive unit tests in `test_tool_component.py`, validating the conversion and functionality of components as tools, including testing various data types and nested structures.

### How can it be used:
- Convert a component to a tool:
  ```python
  tool = Tool.from_component(
      component=myComponent,
      name="my_tool",
      description="A tool for processing data"
  )
  ```
and the use it as a Tool following the established patterns. 

### How did you test it:
- Conducted unit tests covering various component types such as simple return types, data classes, and Pydantic models.
- Performed integration tests with pipelines utilizing both OpenAI and Anthropic backend models, verifying tool correctness and error management.

### Notes for the reviewer:
- Pay attention to the handling of nested data structures and nullable types to ensure they are mapped and invoked correctly.


haystack_experimental/dataclasses/tool.py
```diff
@@ -198,6 +203,66 @@ def get_weather(
...
+            )
+
+        if getattr(component, "__haystack_added_to_pipeline__", None):
+            msg = (
+                "Component has been added in a Pipeline and can't be used to create a Tool. "
+                "Create Tool from a non-pipeline component instead."
+            )
+            raise ValueError(msg)
+
+        from haystack_experimental.components.tools.openai.component_caller import extract_component_parameters
```
anakin87, line 230: I have some doubts related to this line:
Why is this import placed here specifically?
Why is it in `[...].tools.openai.component_caller`? Is it compatible with OpenAI only or is it more generic?

haystack_experimental/dataclasses/tool.py
```diff
@@ -198,6 +203,66 @@ def get_weather(
...
+        if not isinstance(component, Component):
+            raise ValueError(
+                f"{component} is not a Haystack component!" "Can only create a Tool from a Haystack component instance."
+            )
+
+        if getattr(component, "__haystack_added_to_pipeline__", None):
+            msg = (
+                "Component has been added in a Pipeline and can't be used to create a Tool. "
+                "Create Tool from a non-pipeline component instead."
+            )
```
anakin87, line 227: Can you please explain this?

If I remember correctly, one of the requirements was about deserializing Tools from YAML (which should be feasible if Tools are components). I'm not totally sure...

anakin87 (commented):
Thanks for the effort you are putting into this PR.

I do have some questions and suggestions.

The requirement was to support **components with basic str or native Python types as input**. This implementation appears to go beyond that, which is great for flexibility but might become hard to maintain.
- What's the advantage of supporting Pydantic models here? Maybe I am just missing some reasonable use cases...

As I commented in the PR, if I remember correctly, one of the initial requirements was to enable the **deserialization of Tools from YAML**, which would be feasible if Tools are treated as components. Is it possible? If yes, can we add some tests to cover this?

Given the main goal of this PR, would it make sense to involve someone from the DC team in the review?

haystack_experimental/dataclasses/tool.py
vblagoje: Yes, good point. I'll remove openai although one could argue this format is openai while Cohere for example has different format for tool calling definitions. Now, should we leave it in tools namespace with tool_invoker.py or perhaps move it perhaps close to tool.py in the same dataclasses module? The reason why I had it inside the method is circular deps which go away if we collocate `extract_component_parameters` next to tool.py in dataclasses. wdyt?

haystack_experimental/dataclasses/tool.py
vblagoje: Yes, I thought we can have a component declared but not be part of the pipeline. Maybe not, depending on that we can remove this check.

vblagoje:
> Thanks for the effort you are putting into this PR.
> 
> I do have some questions and suggestions.
> 
> The requirement was to support **components with basic str or native Python types as input**. This implementation appears to go beyond that, which is great for flexibility but might become hard to maintain.
> 
> * What's the advantage of supporting Pydantic models here? Maybe I am just missing some reasonable use cases...
> 
The TypeAdapter from pydantic takes care of range of json to object conversion be these objects strings or other native Python types, dataclasses or Pydantic objects. 

It is in fact, harder and longer codebase to implement a version of Tool.from_component that only supports strings, native Python types and their lists as we need to detect cases we don't support, raise errors and so on.

TypeAdapter.validate_python takes conversion of all of these. Unit and integration tests for dataclasses and Pydantic models I included to showcase how this is indeed possible with almost no code. We even support our own Document class as show in test examples.  

Given these newfound benefits of TypeAdapter which enable proper conversion support with 10 lines of well tested pydantic code - I went for that solution although it was not required by the requirements. 

> As I commented in the PR, if I remember correctly, one of the initial requirements was to enable the **deserialization of Tools from YAML**, which would be feasible if Tools are treated as components. Is it possible? If yes, can we add some tests to cover this?
> 

Yes, I need to review this part as well.

> Given the main goal of this PR, would it make sense to involve someone from the DC team in the review?
Yes, @mathislucka is on PTO and he had the most context. Let's wait for him

haystack_experimental/dataclasses/tool.py
anakin87: I still don't understand if this is a self-imposed limitation (I don't think so) or there are strong reasons to avoid that. Could you please explain this point further?

haystack_experimental/tools/__init__.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from .component_schema import create_tool_parameters_schema
+
+__all__ = ["create_tool_parameters_schema"]
```
anakin87, line 7: - I would not export this function here if possible - see https://github.com/deepset-ai/haystack/issues/8650
- I would prefer to make this method internal and also all others in component_schema.py. They should not be user-facing and if we make them internal, we are then free to change them at any time if needed.

anakin87 (commented):
I like the simplification you have made.

I left other comments and asked Julian to take a look as well.

haystack_experimental/tools/__init__.py
vblagoje: Ok makes sense, will do 🙏

haystack_experimental/dataclasses/tool.py
```diff
@@ -198,6 +205,59 @@ def get_weather(
 
         return Tool(name=name or function.__name__, description=tool_description, parameters=schema, function=function)
 
+    @classmethod
+    def from_component(cls, component: Component, name: str, description: str) -> "Tool":
+        """
+        Create a Tool instance from a Haystack component.
+
+        :param component: The Haystack component to be converted into a Tool.
+        :param name: Name for the tool.
```
julian-risch, line 214: I don't want to make this PR more complex but have we considered using defaults for name and description? For example the component name as the default for `name` and parts of the component docstring extracted via docstring_parser for `description`? 
That would also enable the ToolInvoker to call from_component internally if the input to its `tool` parameter is a list of components instead of tools, which I think is needed for defining tools as components in a yaml.

julian-risch (changes requested):
Tool.from_component functionality is great! I have some feedback about the tests and suggest we add defaults for tool name and tool description so that the from_component requires only a component as input for standard use.

For the tests:
I suggest replacing gpt-4o with gpt-4o-mini if that works. Also consider replacing claude-3-5-sonnet-20240620 with a simpler model if that works?

I understand that the long instructions like `"Concatenate these documents: First one says 'Hello world' and second one says 'Goodbye world'. Set only content field of the document only. Do not set id, meta, score, embedding, sparse_embedding, dataframe, blob fields."` are meant to test the implementation. Still, I suggest that we use Document instances as input directly too. For example, have one e2e test with a retriever as a tool from component and another component that takes documents as input as a second tool.
As the test for OpenAI and Anthropic are very similar, let's use parametrization for the tests, which will reduce lines of code. You might need to use `pytest.skip()` within the test to skip if api keys are not available then.
Last but not least, we should have a test case that covers loading tools from components defined in a pipeline yaml. That should be a YAML where a ToolInvoker is defined and it's init parameter for tools is a list of components. The ToolInvoker would need to call from_component internally then as we discussed offline.

vblagoje:
Abandoned and superseded by https://github.com/deepset-ai/haystack/pull/8693

PR #162, opened 2025-01-03
Auto merging retriever update

davidsbatista:
### Related Issues

- fixes #issue-number

### Checklist

- I documented my code


haystack_experimental/components/retrievers/auto_merging_retriever.py
```diff
@@ -117,41 +117,58 @@ def run(self, matched_leaf_documents: List[Document]):
...
-        Groups the matched leaf documents by their parent documents and returns the parent documents if the number of
-        matched leaf documents below the same parent is higher than the defined threshold. Otherwise, returns the
-        matched leaf documents.
+        Recursively groups documents by their parents and merges them if they meet the threshold,
+        continuing up the hierarchy until no more merges are possible.
 
         :param matched_leaf_documents: List of leaf documents that were matched by a retriever
         :returns:
-            List of parent documents or matched leaf documents based on the threshold value
+            List of documents (could be a mix of different hierarchy levels) based on threshold values
```
julian-risch, line 125: There is still only a single threshold value, right? I wonder why it's `threshold values` here then? Are you referring to merge scores here?

haystack_experimental/components/retrievers/auto_merging_retriever.py
```diff
@@ -117,41 +117,58 @@ def run(self, matched_leaf_documents: List[Document]):
...
+                parent = parent_doc[0]
+                if not parent.meta.get("__children_ids"):
+                    raise ValueError(f"Parent document with id {doc_id} does not have any children.")
+
+                # Calculate merge score
+                score = len(child_docs) / len(parent.meta["__children_ids"])
+                if score > self.threshold:
+                    merged_docs.append(parent)  # Merge into parent
+                else:
+                    merged_docs.extend(child_docs)  # Keep children separate
```
julian-risch, line 163: I think we could extend docs_to_return here instead. So `docs_to_return.extend(child_docs)`

haystack_experimental/components/retrievers/auto_merging_retriever.py
```diff
@@ -117,41 +117,58 @@ def run(self, matched_leaf_documents: List[Document]):
...
+
+            # if no new merges were made, we're done
+            if len(merged_docs) == len(documents):
+                return merged_docs
+
+            # Recursively try to merge the next level
+            return try_merge_level(merged_docs + docs_to_return)
```
julian-risch, line 170: I think we could do `return docs_to_return + try_merge_level(merged_docs)` and thereby circumvent the extra work I mentioned.

julian-risch (changes requested):
I have one change request about the `try_merge_level`. Other than that the PR looks good to me.
There are cases where we check multiple times whether we can merge the same `child_docs` over and over in `try_merge_level`.
Imagine we start `try_merge_level` with 4 documents. 
2 documents from a parent A and two from a parent B. Let's say when we calculate merge scores, documents from parent A get merged. documents from parent B don't get merged.
Now, there were some merges made, so we're not done yet. We go into the recursion and check the documents for B again. Let's prevent this from happening. I suggest we change two lines for this but haven't tried it. Please check or maybe you have another idea.

haystack_experimental/components/retrievers/auto_merging_retriever.py
julian-risch: However, it's not "tail recursion" then anymore. If we wanted to keep tail recursion, I would suggest passing two lists to try_merge_level. One list that contains the docs_to_return and that need no further checks. That list is empty at the very beginning. And then there should be a second list with the docs that still need to be checked for merging.

julian-risch:
The class docstring also needs to be updated.
When I run the example code, i get three levels of documents now.
```python
from haystack import Document
from haystack_experimental.components.splitters import HierarchicalDocumentSplitter
from haystack_experimental.components.retrievers.auto_merging_retriever import AutoMergingRetriever
from haystack.document_stores.in_memory import InMemoryDocumentStore

# create a hierarchical document structure with 2 levels, where the parent document has 3 children
text = "The sun rose early in the morning. It cast a warm glow over the trees. Birds began to sing."
original_document = Document(content=text)
builder = HierarchicalDocumentSplitter(block_sizes=[10, 3], split_overlap=0, split_by="word")
docs = builder.run([original_document])["documents"]


>>> len([d for d in docs if d.meta["__level"]==0])
1
>>> len([d for d in docs if d.meta["__level"]==1])
2
>>> len([d for d in docs if d.meta["__level"]==2])
7
```
which contradicts the comment `# create a hierarchical document structure with 2 levels, where the parent document has 3 children`

haystack_experimental/components/retrievers/auto_merging_retriever.py
davidsbatista: typo - I've just removed that last part, not need for it

haystack_experimental/components/retrievers/auto_merging_retriever.py
davidsbatista: this seems to be working :)

haystack_experimental/components/retrievers/auto_merging_retriever.py
davidsbatista: this is yet another good improvement, thanks 👍🏽

julian-risch (changes requested):
Found one more line that needs to be changed in my opinion. As all tests pass already, I would suggest adding a test to cover that change too.

haystack_experimental/components/retrievers/auto_merging_retriever.py
```diff
@@ -117,41 +117,58 @@ def run(self, matched_leaf_documents: List[Document]):
...
+
+            for doc in documents:
+                if doc.meta.get("__parent_id"):  # only docs that have parents
+                    parent_documents[doc.meta["__parent_id"]].append(doc)
+                else:
+                    docs_to_return.append(doc)  # keep docs that have no parents
+
+            # If no documents have parents, we're done
+            if not parent_documents:
+                return documents
```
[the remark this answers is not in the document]
davidsbatista: I can just remove that line, and let the recursion exit in its normal way, that is no more recursions to do - otherwise this was returning duplicates for this specific case

PR #164, opened 2025-01-08
Tool from Component YAML serde POC

anakin87:
## Features to implement

- Enable conversion of components to Tools (essentially implemented by Vladimir in https://github.com/deepset-ai/haystack-experimental/pull/159)
- Support the YAML serialization and deserialization for components defined from Tools

## Idea

1. `ComponentTool` (subclass of `Tool`)
- This would have the same features of `Tool.from_component`.
- `to_dict` method  would handle both the serialization of the Tool itself and the underlying component.
- `from_dict` method would recreate the component and the Tool properly.
- The component would be created in isolation - outside of a Pipeline.

2. Introducing a `type` attribute for `Tool`
As discussed, the lack of a `type` attribute in Tool is problematic for custom serialization/deserialization, especially when deviating from the basic Tool implementation: in Tool from component, Tool from OpenAPI... 
To address this, I propose adding a `type` attribute.

*The implementation is pretty simple and you can see it in action in the example.py file.*

**Pros**
- No changes to the Pipeline serde logic to accommodate this use case.
- Avoids adding potentially confusing methods like `add_tool` to the Pipeline.
- Minimal divergence from the current implementation.
- Clearly separates components wrapped as Tools from the Pipeline: they are independent, cannot be connected, etc.

**Cons**
- not specific to this idea, but we deserialize and initialize the Tools each time they appear (e.g., duplication between `Generator` and `ToolInvoker`). This isn't a significant blocker now but might become problematic if users start converting heavy components into Tools.

## About verbosity/YAML duplication
As agreed, this has a lower priority, but I found something to explore about YAML anchors and aliases: https://ttl255.com/yaml-anchors-and-aliases-and-how-to-disable-them/

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,0 +1,155 @@
+from haystack import Pipeline
+from haystack.components.retrievers.in_memory import InMemoryBM25Retriever
+from haystack.dataclasses import Document
+from haystack.document_stores.in_memory import InMemoryDocumentStore
+
+from haystack_experimental.components.generators.chat import OpenAIChatGenerator
+from haystack_experimental.components.tools import ToolInvoker
+from haystack_experimental.dataclasses import ChatMessage
+from haystack_experimental.dataclasses.tool import ToolComponent
+
+ds = InMemoryDocumentStore()
+
+ds.write_documents(
+    [
+        Document(content="Hello, how can I help you?", meta={"id": "1"}),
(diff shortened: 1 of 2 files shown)


anakin87:
Closing. We're implementing the POC in the linked PRs.

PR #171, opened 2025-01-20
fix: cyclic pipelines should run stable

mathislucka:
### Related Issues

- PR mirrors https://github.com/deepset-ai/haystack/pull/8707

### Notes for the reviewer
Note that the PR does not add any new code in `base.py`. Check the original branch to see the diff. The PR only removes code from `base.py`.

Also note, that the AsyncPipeline will not run anymore when this change is merged to haystack main as it relies on some methods in PipelineBase that I removed.

### Checklist

- I documented my code


davidsbatista:
@mathislucka I think this PR should have at least one more reviewer, ideally +2, this is a big change

mathislucka:
> @mathislucka I think this PR should have at least one more reviewer, ideally +2, this is a big change

Yes, @Amnah199 already started reviewing the PR in Haystack (core). @julian-risch will also review.

davidsbatista:
we need to add `Pipeline` to the correct `__init__.py` hierarchy so that after the release users can simply do:

`from haystack_experimental import Pipeline`

I've been using it to test several pipelines and always have to do a long import, i.e.:

`from haystack_experimental.core.pipeline.pipeline import Pipeline`

pyproject.toml
```diff
@@ -63,6 +63,7 @@ extra-dependencies = [
   # LLMMetadataExtractor dependencies
   "amazon-bedrock-haystack>=1.1.1",
   "google-vertex-haystack>=2.0.0",
+  "ipython"
```
davidsbatista, line 66: I believe this was added for testing/debugging purposes? we should also remove it before the release

mathislucka: No, it was added because the run tests require it (they test pipeline drawing which detects if the pipeline runs in a notebook or in a script.)

davidsbatista: can you add a simple comment on top so that we know why it's needed/who uses it

davidsbatista:
regarding the test files - I tried to run them but it failed due to a mixing fixture - `fixture 'spying_tracer' not found` - then later, talking with @julian-risch, I understood that we want to keep the behavioral tests out of this repo (at least for now). 

I would suggest we completely remove them or add them in total/working - just to keep the PR in a clean state - we can then later add the full tests in another PR - so that if we need to rollback for whatever reason things are in a clean state.

mathislucka:
I found another bug that affects non-cyclic pipelines while working on super components.

Consider the pipeline that is created here: https://github.com/deepset-ai/haystack-experimental/blob/72528c51e158bb2b1ac0e34f6a3bf64d8b74dbd1/haystack_experimental/super_components/converters/file_converter.py#L88

And have a look at the example notebook for the `AutoFileConverter`.

I found that depending on the insertion and connection order, the `tabular_joiner` might run with partial inputs which leads to a change in results.

The `tabular_joiner` will run twice in these cases and it will only ever output either the documents coming from `joiner` or from the csv converter.

This happens because after running all the file converters, the `tabular_joiner` can be the only component in the run queue. Once we pop it from the run queue it will be empty.

This is a problem because: https://github.com/deepset-ai/haystack/blob/c989d9c483a4fc74147e17065779fd6103f4a084/haystack/core/pipeline/pipeline.py#L429

Will check the empty run queue (`all([])`) which is `True`. The component will then be executed, removing (for example) the output from the csv converter from its inputs. This will lead to incomplete inputs when it runs again.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -388,6 +388,37 @@ def _is_queue_stale(priority_queue: FIFOPriorityQueue) -> bool:
         """
         return len(priority_queue) == 0 or priority_queue.peek()[0] > ComponentPriority.READY
 
+    @staticmethod
+    def _is_queue_blocked(pq: FIFOPriorityQueue) -> bool:
```
mathislucka, line 334: @Amnah199 we don't need this method. The priority_queue is always sorted by priority.
That means that if the first element in the queue is BLOCKED, all other elements are blocked too.

You can just do:

```python
candidate = priority_queue.peek()
if candidate is not None and candidate[0] == ComponentPriority.BLOCKED:
  raise ...
```

No need to do any copies then either.

Amnah199: Makes sense. Will update this.

julian-risch (approved):
Looks good to me! 
We tested many different pipelines and will continue with comparing traces of pipelines running with the old and the new logic after this PR is merged. We're checking for any differences in pipeline execution order, changes in component inputs/outputs, and decreases/increases in speed.
In addition, we're still testing the new logic with a custom component that is greedy variadic and has an optional input as described [here](https://github.com/deepset-ai/haystack/pull/8707#discussion_r1924032535).

PR #174, opened 2025-01-22
feat: add SuperComponent

tstadel:
### Related Issues

# SuperComponents

Supercomponents in general behave like any other component. They have init params, from_dict() and to_dict() methods as usual. The init params typically determine how the internal pipeline is constructed (e.g. which components are used).

```python
from haystack_experimental.super_components.converters.file_converter import AutoFileConverter
file_converter = AutoFileConverter()
file_converter.to_dict()
```

## Expanding SuperComponents
What makes SuperComponents special is the ability to expand it by calling their `to_super_component_dict()` method. This converts the component to a generic `SuperComponent` that contains the pipeline constructed by the SuperComponent. From there on the pipeline can be changed in any way.

```python
file_converter.to_super_component_dict()
```

Load as SuperComponent:
```python
from haystack_experimental.core.super_component import SuperComponent

SuperComponent.from_dict(file_converter.to_super_component_dict())
```

### Notes for the reviewer
Check out the super_components.ipynb to get an idea how it's working

### Checklist

- I documented my code


mathislucka:
@tstadel I've updated the PR and did some clean up. Here is what I would propose:

**Naming**
We rename the `PipelineWrapper` to `SuperComponent`.
`SuperComponent` diverges from our normal naming scheme (`*er`) on purpose to signal that this is a different type of component with special features.
I also moved all `SuperComponent`-related changes to `core` for the same reason.

**Inheritance**
We need a slight tweak to how we create ready-made super components (e.g. `AutoFileConverter`).
I decided to create a `SuperComponentBase`-class that any custom super components have to inherit from.
`SuperComponent` inherits from the base class too.
The reason is that all custom super components would have to define `to_dict` and `from_dict` methods if they inherited directly from `SuperComponent`. However, if the custom super components do not have any complex parameters to serialize, they would not have to define these methods because we can handle them with the default serialization in the pipeline.
If a custom super component inherited from `SuperComponent` and would not override the serialization methods, this would lead to confusing situations where it would serialize to a `SuperComponent` in a pipeline instead of the actual type of the custom super component.

Inheriting from `SuperComponentBase` avoids this scenario.


**Further Examples**

I updated the example notebook and made the initialization of the `AutoFileConverter` less complex, so that users can understand it better. It's more verbose but easier to customize and adapt.

Here's another simple super component as an example:

```python
from typing import Optional, Dict, Any

from haystack import component, Pipeline

from haystack.components.generators.chat import AzureOpenAIChatGenerator, OpenAIChatGenerator, HuggingFaceAPIChatGenerator
from haystack.utils import Secret

from haystack_experimental.core.super_component import SuperComponentBase

_GENERATOR_PROVIDERS = {
    "openai": OpenAIChatGenerator,
    "azure": AzureOpenAIChatGenerator,
    "huggingface": HuggingFaceAPIChatGenerator,
}

@component
class AutoGenerator(SuperComponentBase):
    def __init__(
            self,
            model: str = "openai:gpt-4o",
            api_key=Secret.from_env_var("LLM_API_KEY"),
            generation_kwargs: Optional[Dict[str, Any]] = None
    ) -> None:
        if ":" not in model:
            raise ValueError("Model string must be in format 'provider:model_name'")

        provider, model_name = model.split(":")
        if provider not in _GENERATOR_PROVIDERS:
            raise ValueError(
                f"Provider must be one of {list(_GENERATOR_PROVIDERS.keys())}"
            )
        
        generator = _GENERATOR_PROVIDERS[provider](
            api_key=api_key,
            model_name=model_name,
            generation_kwargs=generation_kwargs,
        )
        
        pp = Pipeline()
        pp.add_component("generator", generator)

        super(AutoGenerator, self).__init__(pipeline=pp)
```

mathislucka:
### Benefits

An abstraction layer that wraps around pipelines was requested by the community many times (e.g. https://github.com/deepset-ai/haystack/issues/7638).

They allow to encapsulate an entire pipeline so that it is easier to organize code and keep an overview over very large pipelines.

Additionally, we got feedback that our components are very atomic (which is generally good) and they can be hard to understand by novice users.

`SuperComponentBase` allows us to carefully introduce higher-level abstractions for common component patterns in Haystack, making it easier to get started for novice users.

At the same time, we are not hiding how these components are implemented. Any advanced Haystack user who understands how to build pipelines can look at the code for our super components and customize it if needed.

For deepset Studio, we are giving users the possibility to `zoom-in` to any super component if they want to understand how it works. By offering the `to_super_component_dict`-method, Studio users can decide to convert any super component to a generic `SuperComponent` which will allow them to customize the underlying pipeline.

haystack_experimental/super_components/indexers/document_indexer.py
```diff
@@ -2,88 +2,94 @@
...
         self,
-        embedder: Component,
-        document_store: DocumentStore,
+        model: Optional[str] = None,
+        prefix: str = "",
+        suffix: str = "",
+        batch_size: int = 32,
+        embedding_separator: str = "\n",
+        meta_fields_to_embed: Optional[List[str]] = None,
+        document_store: Optional[DocumentStore] = None,
```
tstadel, line 52: Let's not pass a document_store here. This is just for showcasing, let's use InMemoryDocumentStore hard-coded­ for now.

mathislucka:
I've split this into 2 PRs so that we can discuss the basic SuperComponent abstractions and specific super component implementations separately:

- https://github.com/deepset-ai/haystack-experimental/pull/183
- https://github.com/deepset-ai/haystack-experimental/pull/184

mathislucka:
Closing this in favor of:

https://github.com/deepset-ai/haystack-experimental/pull/183
https://github.com/deepset-ai/haystack-experimental/pull/184

PR #175, opened 2025-01-24
feat: Introduce Agent

mathislucka:
## Agent

This PR introduces an `Agent` component that slots right into Haystack's pipelines.

An Agent consists of an LLM and tools. The LLM repeatedly call tools until a handoff-condition is met. By default, the `Agent` returns when the LLM produces a pure text message without any tool calls.

You run the `Agent` by passing a list of chat messages. The agent returns a list of chat messages including the full sequence of tool calls and tool results.

Under the hood, the `Agent` uses a cyclic pipeline with the following components:
- `ChatGenerator`
- `ToolInvoker`
- `ConditionalRouter`
- `BranchJoiner`

### Usage Example

_This is a simple Agent that can perform web search to answer a user's question._
```python
from haystack.components.websearch import SearchApiWebSearch
from haystack.tools import ComponentTool
from haystack.dataclasses import ChatMessage

from haystack_experimental.components.agents import Agent

search_tool = ComponentTool(
  component=SearchApiWebSearch(),
  name="web_search",
  description="Use this tool to perform a search on the web."
)

system_prompt = """
Answer the users question.
Don't rely on your own knowledge, use the `web_search` utility to search for information on the web.
Break down your searches into multiple queries if you have to. Pass them to `web_search` one by one.
  """

agent = Agent(model="openai:gpt-4o", tools=[search_tool], system_prompt=system_prompt)

result = agent.run(messages=[ChatMessage.from_user("Give me the Oscar winners for best actor for the past 3 years and summarize the plot of the films that they won with.")]

# The agent will run queries repeatedly until it returns a response.
# Access the response and the full history of tool calls and tool results
for msg in result["messages"]:
  print(msg)
```

### handoff
Currently, the `Agent` exits the loop when the `handoff`-condition is met.
The default is `Agent(handoff="text")` which means that the agent exits the loop when it produces a text-only response.
Users can set `handoff` to the name of any tool, which will cause the agent to return, after that tool was called.
For the web search example, if we set `handoff="web_search"`, the agent would return after the `web_search`-tool was called and executed once.

_Extending 'handoff'_
We could extend `handoff` and allow users to pass a Jinja2-expression in addition to tool names and "text".
The `ConditionalRouter` has access to the message history, the last message generated from the `ChatGenerator` and the last tool result message from the `ToolInvoker`. Let's say we wanted to return once the `ChatGenerator` generates a message that contains the word 'Done':

`Agent(handoff="{{ 'Done' in llm_messages[0].text }}")`

This isn't implemented yet, but we could do it.

### Passing additional context to tools
In many cases, it is useful to pass additional inputs to a tool or to return results from a tool directly to the pipeline.
Assume we want to add a list of all the search results that the agent viewed to construct an answer with sources.
Additionally, we want to restrict our web search agent to only search on Wikipedia.

This can be achieved by defining `input_variables` and `output_variables` when initializing the `Agent` and by defining tools that accept `ctx`.

```python
from haystack.components.websearch import SearchApiWebSearch
from haystack.tools import ComponentTool
from haystack.dataclasses import ChatMessage
from haystack import Pipeline
from haystack.components.builders import PromptBuilder

from haystack_experimental.components.agents import Agent
from haystack_experimental.tools.tool_context import ToolContext


def web_search(ctx: ToolContext, query: str):
  # we can get additional inputs from the 'ctx' by calling get_input
  allowed_domains = ctx.get_input("allowed_domains")
  comp = SearchApiWebSearch(allowed_domains=allowed_domains)
  
  result = comp.run(query=query)
  
  if existing_docs := cxt.get_output("documents"):
    # we can pass on additional outputs from the tool by calling set_output
    ctx.set_output("documents", existing_docs + result["documents"])
  else:
    ctx.set_output("documents", result["documents"])

  return result["documents"]


system_prompt = """
Answer the users question.
Don't rely on your own knowledge, use the `web_search` utility to search for information on the web.
Break down your searches into multiple queries if you have to. Pass them to `web_search` one by one.
  """

agent = Agent(
  model="openai:gpt-4o",
  tools=[search_tool],
  system_prompt=system_prompt,
  input_variables={"allowed_domains": List[str]},
  output_variables={"documents": List[Document]}
)

# let's use a PromptBuilder to get a response in a nice format
answer_template = """
{{ (messages|last).text }}

_Viewed Search Results:_
{% for doc in documents %}
{{ doc.content }}
{% endfor %}
"""

answer_formatter = PromptBuilder(template=answer_template)

pipe = Pipeline()

pipe.add_component("agent", agent)
pipe.add_component("answer", answer_formatter)

pipe.connect("agent.messages", "answer.messages")

# note that the agent now has an output socket for documents
pipe.connect("agent.documents", "answer.documents")

user_input = [ChatMessage.from_user("Give me the Oscar winners for best actor for the past 3 years and summarize the plot of the films that they won with.")]

# Note how the agent now has an "allowed_domains" input socket.
result = pipe.run(data={"agent": {"messages": user_input, "allowed_domains": ["wikipedia.org"]}})
```

Apart from passing additional inputs to tools and getting additional outputs from the agent, we can also pass values from one tool to another by writing to and reading from `ToolContext.outputs`.

_To be discussed:_

- This mechanism does not work with `ComponentTool` yet. We should discuss how the same behaviour could be implemented there.

- I opted for a simple dataclass used as `ToolContext`. This dataclass does not perform type checking at runtime. Theoretically, we could implement a version of `ToolContext` using Pydantic dynamic models, that would validate types at runtime but it's a more complex implementation, brings reliance on Pydantic and I think it has limited benefits.

### Tracing

Tracing generally works with all of our different tracing integrations. However, since the `Agent` uses a pipeline internally, it would be great if we could associate that trace with the pipeline that called the agent. This is a general enhancement that would also be useful for `SuperComponent` as [implemented here](https://github.com/deepset-ai/haystack-experimental/pull/174). I think we should tackle it in a separate PR though.

### Checklist

- I documented my code


anakin87:
I want to share some initial thoughts and questions, mostly at a high level.
It would be great if @vblagoje could also review this PR and if @julian-risch and @bilgeyucel could take a look...

### Agent as a component
- I would like to understand the reasoning behind making `Agent` a `component`.
My guess is that this might be to integrate it into Pipelines, but it seems a bit different from the principle we’ve followed in the past (keeping components small and self-contained).  Could you share what you see as the pros and cons of this approach?

### `ToolContext`
- I noticed that `ToolContext` is heavily used in this implementation. While I understand it can sometimes be useful, I wonder if there’s an opportunity to simplify things. Do you think there’s room to reduce its usage or rethink other classes to make it less central?
I would also love to hear what others think about this.

- Slightly related. The problem mentioned in the PR regarding `ComponentTool`: is it specifically tied to `ToolContext`, or does it also involve `Agent`?

### Implementation details
1. I noticed that Agent is implemented as a Pipeline instead of plain Python, which seems to make the code longer and more complex. Since this is an internal implementation, is there a specific reason for choosing Pipelines here?

2. In OpenAI Swarm, `handoff` is typically used to describe passing control to another Agent, but here it seems to indicate that the Agent is exiting. Would it make sense to consider alternatives like `return_condition` or `end_condition`?

3. (not directly related to this PR)  I’ve noticed for a while now that we don’t have a unified way to qualify `ChatGenerators`. I would avoid abstract classes, but maybe a thin `Protocol` could solve this.

(Could you rebase this PR to make it smaller and easier to review?)

mathislucka:
Thanks for the initial feedback!


> I would like to understand the reasoning behind making Agent a component.
My guess is that this might be to integrate it into Pipelines, but it seems a bit different from the principle we’ve followed in the past (keeping components small and self-contained). Could you share what you see as the pros and cons of this approach?

Yes, exactly, this allows us to integrate it into pipelines. See the notebook for an example of how using it in a pipeline is beneficial.

Pros:
- combine agentic and non-agentic workflow steps
- users don't need to learn a new abstraction, they are already familiar with pipelines and components
- use pipelines to orchestrate multi-agent systems
- we get tracing and debugging utilities already built into pipelines for free

Cons:
I don't really see any. @vblagoje brought up initially that it would be great if users wouldn't have to know that it is a component but that's already the case. Users don't need to know anything about components or pipelines to call `Agent.run(messages)`. We could even allow them to `from Haystack import Agent` if we wanted to make it more prominent.

 

> I noticed that ToolContext is heavily used in this implementation. While I understand it can sometimes be useful, I wonder if there’s an opportunity to simplify things. Do you think there’s room to reduce its usage or rethink other classes to make it less central?
I would also love to hear what others think about this.

I tried to keep the use of ToolContext light. It's only used in two places. For users implementing tools on the other hand, ToolContext becomes more central and they need to understand it to use it. We can keep things easy though because ToolContext is optional, so if users don't need it, they don't even need to know it exists.

I found the possibility to pass additional context to tools from outside the agent, as well as passing context between tools, and passing outputs from tools to downstream components in the pipeline very useful when implementing agents.

I'm not entirely happy with how ToolContext needs to be used by users implementing tools though. I'd be very happy about alternative suggestions.

> Slightly related. The problem mentioned in the PR regarding ComponentTool: is it specifically tied to ToolContext, or does it also involve Agent?

ComponentTool can be used with Agent without any problems. The only thing that I didn't implement is support for ToolContext. I think that for additional inputs it would be fairly easy to implement. We could just check the input sockets of the components and pass any `ToolContext.inputs` that match the input sockets to the component. This would also remove the burden from the user to implement handling of additional inputs. For outputs however, this is more difficult. Right now, I'm delegating the responsibility for how to handle existing outputs from the same or other tools to the user who implements the function for the tool. In the Issue Agent example in the notebook, I'm extending a list of documents with every tool call.

Assume we move that responsibility to `ComponentTool.invoke`:
We could check the component outputs for any matches to ToolContext.outputs. However, there would be open questions for me with regards to how to handle these outputs:
- overwrite existing outputs?
- extend them?
- are outputs that went into ToolContext.outputs excluded from the ToolCallResult?

Ideally, I'd want to leave that decision to the user but I'm not sure how.

 

> I noticed that Agent is implemented as a Pipeline instead of plain Python, which seems to make the code longer and more complex. Since this is an internal implementation, is there a specific reason for choosing Pipelines here?

I agree that using a pipeline is more verbose but I wouldn't say it's more complex. My main reasons were:
- Haystack already foresees to use components for agentic behaviour (ChatGenerator, ToolInvoker) and the handoff condition was easy to implement with the ConditionalRouter, so since I was already using 3 components, it seemed natural to connect them using a pipeline
- we get tracing for free
- we can extend the implementation easily by simply adding components
- users can take the implementation, integrate it directly into their pipelines and extend if needed

 

> In OpenAI Swarm, handoff is typically used to describe passing control to another Agent, but here it seems to indicate that the Agent is exiting. Would it make sense to consider alternatives like return_condition or end_condition?

Yes, I know. I interpreted handoff as handing off control of the execution flow, so the idea was that the agent basically hands off to the pipeline. This handoff could also mean that it hands off to another Agent, if there is another Agent in the pipeline. However, I see how this might confuse users. What about `exit_condition`?

I merged main into the PR to remove the diff from the pipeline.run refactoring. You can also ignore the github components in the examples folder for now. There are more components than we actually need for the example notebook. My intention was to add a second agent as an example but I didn't get to it yet.

mathislucka:
_added the discussed changes with regards to tool context/state_


*Suggested Approach*

We introduce a simple `State`-class that is used in the following components:

- `Agent` uses it internally
- either the existing `ToolInvoker`  (preferred) accepts state as a `run`-parameter and returns state  or we create a `StateAwareToolInvoker` that is derived from the existing implementation

The `ToolInvoker` handles all reads from and writes to the state. It passes relevant parameters to tools and updates the state with their return values where applicable.

***Read docs for state below***

### **What is State?**

State lets you provide additional inputs to your tools (beyond what the LLM generates) and capture extra outputs from those tools. This is useful when some tool parameters come from outside the LLM (e.g., user credentials) or when you need to accumulate tool results (e.g., a growing list of retrieved documents).

You define a *state schema* to declare which additional attributes you want to manage, along with their types and (optionally) a default handler function.

```python
from typing import List
from haystack import Document

state_schema = {
    "repository": {"type": str},
    "documents": {"type": List[Document]},
}

agent = Agent(
    model="openai:gpt-4o",
    state_schema=state_schema
)

# The Agent now has 'repository' and 'documents' as "extra sockets"
# for inputs/outputs.

```

*Type Mismatch Note*

- If a tool returns data that is incompatible with your declared type, or your handler produces a final value that doesn’t match the type, an error or unexpected behavior may occur. Ensure each attribute’s usage and handler align with its declared type.

---

### **Reading from State (Tool Inputs)**

Tools can automatically receive arguments from the state if their run method has matching parameter names. For instance, if you have `"repository"` declared in your state and your tool’s `run(repository: str, ...)` method, that parameter is filled from the state.

**Explicit Input Mappings**

You can also define a mapping of `{ "<state_key>": "<tool_param>" }` to specify exactly which state attributes go into which tool parameters. Doing so *limits* which attributes are passed to the tool. Only the keys you define in `inputs` are injected:

```python
repo_viewer = GitHubRepositoryViewer()

repo_tool = ComponentTool(
    name="view_repo",
    description="Use to explore GitHub repositories.",
    component=repo_viewer,
    inputs={"repository": "repository"}  # Only 'repository' is passed
)

# For function-based tools:
@tool(inputs={"repository": "repo"})
def view_repository(path: str, repo: str):
    ...

```

**Conflict Resolution**

- If your tool has a parameter that is *not* listed in `inputs`, it will not receive that attribute from state (the LLM must provide it, or it won’t be provided at all).
- If both the LLM and the state specify a value for the same parameter, the LLM’s value takes precedence.

---

### Writing to State (Tool Outputs)

Whenever a tool returns a dictionary of outputs, the agent merges that dictionary into the global state according to “handler” rules. By default, the system infers how to merge based on your declared type:

- **List**: If the returned value is also a list, it’s extended (`list.extend`). Otherwise, a non-list returned value is appended to the state’s list.
- **Primitive** (str, int, bool, float): The new value replaces the old one.
- **Dict**: The returned dict merges into the existing dict (like a shallow `dict.update`).
- **Other**: The existing value is replaced entirely.

**Custom Handlers**

You can override these defaults in two ways:

1. **Globally** in the `state_schema`:
    
    ```python
    def append(current_state, update):
        current_state.append(update)
        return current_state
    
    state_schema = {
        "repository": {"type": str},
        "documents": {"type": List[Document], "handler": append}
    }
    
    agent = Agent(
        model="openai:gpt-4o",
        state_schema=state_schema
    )
    
    ```
    
    In this example, each new output for `"documents"` is appended as a single element, rather than extended.
    
2. **Locally for a specific tool** by specifying `outputs={"<state_key>": {...}}`:
    
    ```python
    def replace_handler(current_state_value, update):
        return update  # Replace the existing value fully
    
    repo_tool = ComponentTool(
        name="view_repo",
        description="Use to explore GitHub repositories.",
        component=repo_viewer,
        inputs={"repository": "repository"},
        outputs={"documents": {"source": "documents", "handler": replace_handler}}
    )
    
    ```
    
    Here, the local override for `"documents"` means the default or global handler is ignored, and `replace_handler` is used whenever **this** tool returns a `"documents"` key.
    

### Precedence

1. **Local (tool-level) outputs** override the global schema handler for the same state key.
2. If no local override is provided, the global schema handler applies.
3. If no global handler is provided, the system uses the default type-based logic.

---

## Custom Message Handling

By default, *all* of a tool’s return values are stringified and appended as a single `ChatMessage` with the role of `TOOL`. This ensures that the conversation logs every result for the LLM to see on the next turn.

### Creating a Custom Chat Message

You can pick which part of the returned dict becomes the tool message via:

```python
def docs_to_str(docs):
    result = ""
    for doc in docs:
        result += f"{doc.content}\n\n"
    return result

repo_tool = ComponentTool(
    name="view_repo",
    description="Use to explore GitHub repositories.",
    component=repo_viewer,
    outputs={
        "message": {
            "source": "documents",      # Only pass 'documents' to the handler
            "handler": docs_to_str      # Format all docs into a single string
        },
        "documents": {
            "source": "documents",      # Meanwhile store them in state
            "handler": None             # or rely on the default or global schema handler
        }
    }
)

```

**How this works**:

- `"message"` is a special output key telling the system, “this is how to create the chat message.”
- If `source` is omitted, the handler receives the *entire* tool return. Otherwise, it receives only `tool_return["documents"]`.
- If you want multiple outputs from one tool (e.g. store `"documents"` in state while also producing a summary as a message), you can define multiple entries in `outputs`.

---

mathislucka:
@sjrl @anakin87

I updated the PR to reflect the `State` changes that we discussed last week.
There are still some minor things missing (mostly related to serialization) but I would love if you could give this PR another pass.

I also decided to split the notebook from this PR to make the diff smaller. You can see the `Agent` in action here: https://github.com/deepset-ai/haystack-experimental/blob/18d7270e87fcd7b16deaa2f39f363a4f4fe918cb/examples/agent.ipynb

https://github.com/deepset-ai/haystack-experimental/pull/204 should be reviewed and merged after this PR.

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+
+    assert "messages" in result  # Contains conversation history
+    assert "total_tokens" in result  # Contains tool execution outputs
+    ```
+    """
+
+    def __init__( # pylint: disable=too-many-positional-arguments
+        self,
+        model: str,
```
sjrl, line 69: I think we mentioned about directly passing a component here (so a ChatGenerator) instead of a string. This would allow us to drop the `generation_kwargs` and the `api_key`. Also this current version would not support the AzureOpenAIChatGenerator, BedrockChatGenerator, or the GeminiChatGenerator which all require more secrets to be passed than just the api key.

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+        Initialize the agent component.
+
+        :param model: Model identifier in the format "provider:model_name"
+        :param generation_kwargs: Keyword arguments for the chat model generator
+        :param tools: List of Tool objects available to the agent
+        :param exit_condition: Either "text" if the agent should return when it generates a message without tool calls
+            or the name of a tool that will cause the agent to return once the tool was executed
+        :param input_variables: Dictionary mapping input variable names to their types
+        :param output_variables: Dictionary mapping output variable names to their types
```
sjrl, line 86: Remove and replace with a `state_schema` doc string

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+
+with LazyImport(message="Run 'pip install anthropic-haystack' to use Anthropic.") as anthropic_import:
+    from haystack_integrations.components.generators.anthropic.chat.chat_generator import (
+        AnthropicChatGenerator,
+    )
+
+_PROVIDER_GENERATOR_MAPPING = {
+    "openai": OpenAIChatGenerator,
+    "anthropic": AnthropicChatGenerator,
+}
```
sjrl, line 27: This needs to be folded into the LazyImport context above otherwise the `AnthropicChatGenerator` type here will cause an error.

sjrl: Or I guess rather 
```suggestion
_PROVIDER_GENERATOR_MAPPING = {
    "openai": OpenAIChatGenerator,
}
if anthropic_import.is_successful():
    _PROVIDER_GENERATOR_MAPPING["anthropic"] = AnthropicChatGenerator
```

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+        if provider == "anthropic":
+            anthropic_import.check()
+
+        # Initialize components
+        generator = _PROVIDER_GENERATOR_MAPPING[provider](
+            model=model_name,
+            tools=self.tools,
+            api_key=self.api_key,
+            generation_kwargs=self.generation_kwargs,
+        )
```
sjrl, line 136: I think it would be good to make sure we add a `warm_up` method to the `Agent` which calls all potentially relevant `warm_up` methods of it's subsequent components. For example, some ChatGenerators could have warm_up methods (e.g. only initialize the client at warm_up time). 

Maybe this is also relevant for tools? Not entirely sure how we handle tools that are Pipelines.

haystack_experimental/components/agents/agent.py
anakin87: I see... It uses the `Component` protocol. In this case, we would need something more specific: a `ChatGenerator` protocol... IDK if we want to include it in this experiment, but is definitively something to explore...

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+        ]
+
+        router = ConditionalRouter(
+            routes=routes,
+            custom_filters={"assistant_message": ChatMessage.from_assistant},
+            unsafe=True,
+        )
+
+        # Set up pipeline
+        self.pipeline = Pipeline()
```
sjrl, line 177: Exposing `max_runs_per_component` in the init of Agent could be nice here especially since we will be doing looping.

haystack_experimental/components/agents/agent.py
sjrl: That's fair. I think we should use a protocol for this in the version of Agents we bring into main. But I'm fine for this experiment to only work with a subset of ChatGenerators.

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+        :param messages: List of chat messages to process
+        :param kwargs: Additional keyword arguments matching the defined input types
+        :return: Dictionary containing messages and outputs matching the defined output types
+        """
+        state = State(schema=self.state_schema, data=kwargs)
+
+        if self.system_prompt is not None:
+            messages = [ChatMessage.from_system(self.system_prompt)] + messages
+
+        self.pipeline.warm_up()
```
sjrl, line 251: I guess we should just move this to the `warm_up` method of `Agent` right?

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+
+        result = self.pipeline.run(
+            data={
+                "joiner": {"value": messages},
+                "context_joiner": {"value": state},
+            },
+            include_outputs_from={"context_joiner"},
+        )
```
sjrl, line 259: @mathislucka just a general question. Let's say I wanted to see the whole "thinking" process of the Agent, so all intermediate outputs before the exit condition is met. Is there an easy way to get this (e.g. creating the right kind of State schema?)

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+    assert "messages" in result  # Contains conversation history
+    assert "total_tokens" in result  # Contains tool execution outputs
+    ```
+    """
+
+    def __init__( # pylint: disable=too-many-positional-arguments
+        self,
+        model: str,
+        tools: Optional[List[Tool]] = None,
+        system_prompt: Optional[str] = None,
```
sjrl, line 71: Don't know if this is a silly idea, but would it make sense to allow users to pass an initial set of `messages`? This would allow maximum flexibility in how to provide instructions to the underlying ChatGenerator. 

E.g. I believe not all LLM providers use a system prompt.

Or maybe a user would like to pass one system message and one user message. E.g. the system prompt contains fixed instructions with no dynamic variables and then the user message takes in all dynamic requests.

sjrl: Hmm okay I see looking more into your notebook you basically do this already by sending dynamic messages at run time through the run method of Agent. In that case I think just system_prompt alone makes sense.

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+        messages=[ChatMessage.from_user("Find information about Haystack")],
+        search_depth=2
+    )
+
+    assert "messages" in result  # Contains conversation history
+    assert "total_tokens" in result  # Contains tool execution outputs
+    ```
+    """
+
+    def __init__( # pylint: disable=too-many-positional-arguments
```
sjrl, line 67: Maybe not now, but we should look at how we could enable streaming for Agents. Will probably need a streaming_callback in the init and run methods like our ChatGenerators

haystack_experimental/components/agents/agent.py
```diff
@@ -0,0 +1,264 @@
...
+                f"Provider must be one of {list(_PROVIDER_GENERATOR_MAPPING.keys())}"
+            )
+
+
+        valid_exits = ["text"] + [tool.name for tool in tools or []]
+        if exit_condition not in valid_exits:
+            raise ValueError(f"Exit condition must be one of {valid_exits}")
+
+        if state_schema is not None:
+            _validate_schema(state_schema)
```
sjrl, line 104: I was looking at your note

> If a tool returns data that is incompatible with your declared type, or your handler produces a final value that doesn’t match the type, an error or unexpected behavior may occur. Ensure each attribute’s usage and handler align with its declared type.

and I was wondering if it would be possible here to validate the state_schema against the output type of the Tools? Could be something for later since I'm unsure if there is an easy way for us to grab the output types of each Tool.

haystack_experimental/components/agents/agent.py
mathislucka: I'll change it and pass in `Union[OpenAIChatGenerator, AnthropicChatGenerator]` for now. Since it's just a type hint people could presumably also pass other Generators. I only tested with these two though.

haystack_experimental/components/agents/agent.py
mathislucka: Yes, I can do that. One thing I noticed about the `max_runs_per_component` while working on the pipeline.run logic is that we raise an exception when we exceed the limit. I'd say that that isn't very useful for an Agent. It would be possible to change the logic and just remove components that reached their limit from the set of components that can be scheduled. This is for another discussion though.

haystack_experimental/components/agents/agent.py
sjrl: Serialization and deserialization would only work for those two, but yeah if just using Haystack code then it could also just work with other chat generators.

haystack_experimental/components/agents/agent.py
mathislucka: The agent returns all chat messages including any intermediate tool calls, tool results, etc.

I also thought about adding the messages to state by default but came to the conclusion that it isn't needed because we already get them as a proper output.

It might be relevant to add them to State when we think about interrupting an agent (breakpoints) and later picking back up from where it left off. I'd say we keep it like it is right now and can still change it later on.

haystack_experimental/components/agents/agent.py
mathislucka: Yes, the user can pass a list of chat messages. The system_prompt is optional and it can also be overridden when the user passes a chat message with role system to the run method.

haystack_experimental/components/agents/agent.py
mathislucka: I just wanted to start implementing this but I hit a roadblock. If we pass in the Generator, we need to define tools twice: on the Agent so that they can be passed to the ToolInvoker and when we initialize the ChatGenerator.

I'd say that this wouldn't help simplicity and could confuse users. For yaml definitions it would be very verbose. Since the main point of this abstraction is to make things simpler, I would like to avoid passing in the initialized ChatGenerator.

anakin87: If we want to explore this path, I think it would be legitimate/feasible to expect generators without tools from users and then assign the tools to the generators at runtime...

mathislucka: Thanks @anakin87 I took your suggestion and am passing tools at runtime now. It works quite well. I've also updated the example.

mathislucka:
Incorporated the feedback so far and:

- moved to passing the `ChatGenerator` via init and passing tools at runtime (as suggested by @anakin87 )
- removed `State` serde for now (we can add it when we actually need to serialize it)
- updated `ComponentTool` serde to the lastest changes
- I distilled all the comments and ideas from @sjrl [here](https://www.notion.so/deepsetai/Agent-1a1e210b37c4805e9363c3b5d7c14eb1?pvs=4#1a1e210b37c480c7be2dd90660cf1ea6)

I also updated the example notebook with the latest changes (the other PR).

From my POV this should be ready for a full review now.

haystack_experimental/tools/component_tool.py
```diff
@@ -86,13 +88,32 @@ class ComponentTool(Tool):
...
+            outputs: Optional[Dict[str, Any]] = None,
+    ):
         """
         Create a Tool instance from a Haystack component.
 
         :param component: The Haystack component to wrap as a tool.
         :param name: Optional name for the tool (defaults to snake_case of component class name).
         :param description: Optional description (defaults to component's docstring).
+        :param parameters:
+            A JSON schema defining the parameters expected by the Tool.
```
sjrl, line 107: could we a statement about what this defaults to if not provided?

haystack_experimental/tools/component_tool.py
```diff
@@ -109,8 +130,9 @@ def __init__(self, component: Component, name: Optional[str] = None, description
             )
             raise ValueError(msg)
 
+        self.unresolved_parameters = parameters
```
sjrl, line 133: Just a thought, is this worth exposing? We could make this a private variable as well if it's only really useful for the `to_dict` method.

PR #180, opened 2025-01-27
feat: AsyncPipeline that can schedule components to run concurrently

mathislucka:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 This implements a new version of the `AsyncPipeline` that works with the new `Pipeline.run`-logic. Additionally, it can run components within the pipeline concurrently.

### How did you test it?
I re-introduced the behavioral tests to test it. The outputs don't change when compared to the synchronous version of the pipeline.

### Notes for the reviewer

Check out the last part of the updated notebook to see how components run concurrently.

### Checklist

- I documented my code


examples/async_pipeline.ipynb
```diff
@@ -18,11 +18,11 @@
     "- Allow individual components to opt into `async` support.\n",
```
davidsbatista, line 18: <p>Code to initialise the document store seems to be missing, something like this:</p><p> </p><p> </p><p> </p><p><code><span class="ql-cursor">﻿</span>from datasets import load_dataset</code></p><p><code>from haystack import Document</code></p><p><code>from haystack.components.embedders import SentenceTransformersDocumentEmbedder</code></p><p><code>from haystack.document_stores.in_memory import InMemoryDocumentStore</code></p><p> </p><p><code>document_store = InMemoryDocumentStore()</code></p><p><code>dataset = load_dataset("bilgeyucel/seven-wonders", split="train")</code></p><p> </p><p><code>docs = [Document(content=doc["content"], meta=doc["meta"]) for doc in dataset]</code></p><p><code>doc_embedder = SentenceTransformersDocumentEmbedder(model="sentence-transformers/all-MiniLM-L6-v2")</code></p><p><code>doc_embedder.warm_up()</code></p><p> </p><p><code>docs_with_embeddings = doc_embedder.run(docs)</code></p><p><code>document_store.write_documents(docs_with_embeddings["documents"])</code></p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepset-ai/haystack-experimental/pull/180/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

test/core/pipeline/features/test_run.py
```diff
@@ -0,0 +1,2611 @@
...
+
+from pytest_bdd import scenarios, given
+import pytest
+
+from haystack import Document, component
+from haystack.document_stores.types import DuplicatePolicy
+from haystack.dataclasses import ChatMessage, GeneratedAnswer
+from haystack.components.routers import ConditionalRouter
+from haystack.components.builders import PromptBuilder, AnswerBuilder, ChatPromptBuilder
+from haystack.components.preprocessors import DocumentCleaner, DocumentSplitter
```
davidsbatista, line 13: `DocumentSplitter` can be removed, it's not being used

davidsbatista:
It seems there are three main changes to look for:

1) Introducing the `AsyncPipeline`
2) Moving some methods from `Pipeline` to `PipelineBase`
3) Adding the Pipeline tests

Regarding the 1. `AsyncPipeline`
- I went through the logic of the AsyncPipeline it seems OK to me, although there are a few details that require more deep analysis.
- I've run the pipeline in the notebook, and it worked asynchronously, the example needs to be updated with the populating of an async InMemoryDocumentStore


Regarding the move from methods from `Pipeline` to `PipelineBase`
Seems to cause no major issues there, essentially all the methods related to the new logic are now part of the `PipelineBase`

Regarding the tests:
- OpenSearch tests are failing, there's [no opensearch instance running in the background](https://github.com/deepset-ai/haystack-experimental/actions/runs/12994544774/job/36239252700?pr=180#step:5:33) - I suspect this failure existed before this PR already
- test/core/pipeline/features/test_run.py::test_running_a_correct_pipeline[that has two loops of different lengths] - the [logs show that the test passed](https://github.com/deepset-ai/haystack-experimental/actions/runs/12994544774/job/36239335354?pr=180#step:5:392) but with a Error: RuntimeError('Event loop is closed'), seems like the Pipeline is trying to run an async task after the event loop has been closed.

mathislucka:
Thanks for the review @davidsbatista !

> OpenSearch tests are failing, there's [no opensearch instance running in the background](https://github.com/deepset-ai/haystack-experimental/actions/runs/12994544774/job/36239252700?pr=180#step:5:33) - I suspect this failure existed before this PR already

The Opensearch tests are being skipped. This was explicitly before this PR in case there is no running Opensearch instance: https://github.com/deepset-ai/haystack-experimental/blob/5614b4915b843edaa5094aaa1ff8fa326bfca161/test/document_stores/opensearch/test_document_store.py#L350

We could add the setup of an Opensearch instance to the CI but it would make it slow and I see limited benefits because we're running a full test suite in core-integrations. Would you agree?

> test/core/pipeline/features/test_run.py::test_running_a_correct_pipeline[that has two loops of different lengths] - the [logs show that the test passed](https://github.com/deepset-ai/haystack-experimental/actions/runs/12994544774/job/36239335354?pr=180#step:5:392) but with a Error: RuntimeError('Event loop is closed'), seems like the Pipeline is trying to run an async task after the event loop has been closed.

I ran a few iterations of the tests and this isn't related to any of the pipeline run tests. The error already occurred before this PR. It first appeared in  this PR: https://github.com/deepset-ai/haystack-experimental/pull/140 but I suspect it was introduced with the OpenAI tests when we added tool support.
The error comes from the httpx client, which is used in the OpenAI package.

You can see that the error doesn't show up anymore when I remove the tests for the openai chat generator. I would ignore the error for this PR because it isn't related to the changes and it's likely caused by something like improper tear down from the testing framework and probably doesn't have any real world impact. What do you think?

examples/async_pipeline.ipynb
mathislucka: This code was already there before the PR. You can see it in the first part of the notebook here: https://github.com/deepset-ai/haystack-experimental/blob/5614b4915b843edaa5094aaa1ff8fa326bfca161/examples/async_pipeline.ipynb

I only added the example for concurrent execution of components within the pipeline at the end of the notebook.

davidsbatista: <p>It wasn't showing up in the diff</p>

davidsbatista:
> We could add the setup of an Opensearch instance to the CI but it would make it slow and I see limited benefits because we're running a full test suite in core-integrations. Would you agree?

Yes, agree.

> You can see that the error doesn't show up anymore when I remove the tests for the openai chat generator. I would ignore the error for this PR because it isn't related to the changes and it's likely caused by something like improper tear down from the testing framework and probably doesn't have any real world impact. What do you think?

okay, fine from my side 👍🏽

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+                instance: Component = comp_dict["instance"]
+                with tracing.tracer.trace(
+                    "haystack.component.run_async", parent_span=parent_span
+                ):
+                    logger.info("Running component {name}", name=component_name)
+
+                    if getattr(instance, "__haystack_supports_async__", False):
+                        outputs = await instance.run_async(**component_inputs)  # type: ignore
+                    else:
+                        loop = asyncio.get_event_loop()
```
Amnah199, line 122: Can you clarify if here the intention is to create another another event loop if it doesn't exist. If not, we can also look into using `asyncio.get_running_loop()` which is recommended to prevent unexpected behaviour. 
Also asyncio.get_event_loop() is deprecated since python 3.12.
[Source](https://docs.python.org/3/library/asyncio-eventloop.html)
```suggestion
                        loop = asyncio.get_running_loop()
```

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+                        )
+
+                comp_dict["visits"] += 1
+
+                if not isinstance(outputs, dict):
+                    raise PipelineRuntimeError(
+                        f"Component '{component_name}' returned non-dict outputs."
+                    )
+
+                # Distribute outputs to downstream inputs; also prune outputs
```
Amnah199, line 134: Just for clarity 
```suggestion
                # Distribute outputs to downstream inputs; also prune outputs based on `include_outputs_from`
```

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+                        # Edge case: somehow became HIGHEST => run in isolation
+                        async for partial_out in _run_highest_in_isolation(component_name):
+                            yield partial_out
+                        return
+
+                    # else it remains DEFER or DEFER_LAST, keep waiting
+                    if running_tasks:
+                        # Wait for just one task to finish
+                        async for part in _wait_for_one_task_to_complete():
+                            yield part
```
Amnah199, line 289: We are re-calculating the priority in every iteration of while loop. If the priority can only change when a task is finished, then we can calculate it once before entering the loop and once in `if running_tasks:` block (after a task completion).  If I misunderstood it, would be nice to know anyways.
```suggestion
                comp_dict = self._get_component_with_graph_metadata(component_name)
                    new_prio = self._calculate_priority(comp_dict, inputs_state.get(component_name, {}))
                while True:
                    # Already scheduled => stop
                    if component_name in scheduled_components:
                        return
                    
                    if new_prio == ComponentPriority.READY:
                        # It's now ready => schedule it
                        await _schedule_ready_task(component_name)
                        return

                    elif new_prio == ComponentPriority.HIGHEST:
                        # Edge case: somehow became HIGHEST => run in isolation
                        async for partial_out in _run_highest_in_isolation(component_name):
                            yield partial_out
                        return

                    # else it remains DEFER or DEFER_LAST, keep waiting
                    if running_tasks:
                        # Wait for just one task to finish
                        async for part in _wait_for_one_task_to_complete():
                            yield part
                          # Recalculate priority only after a task finishes
                          new_prio = self._calculate_priority(comp_dict, inputs_state.get(component_name, {}))
                            
```

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+
+                :param component_name: The name of the component.
+                :returns: An async iterator of partial outputs.
+                """
+                comp_dict = self._get_component_with_graph_metadata(component_name)
+                while True:
+                    # Already scheduled => stop
+                    if component_name in scheduled_components:
+                        return
+                    # Re-check the actual priority now
```
Amnah199, line 272:
```suggestion
                    # Priority is recalculated after each completed task

```

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+                        outputs = await loop.run_in_executor(
+                            None, lambda: instance.run(**component_inputs)
+                        )
+
+                comp_dict["visits"] += 1
+
+                if not isinstance(outputs, dict):
+                    raise PipelineRuntimeError(
+                        f"Component '{component_name}' returned non-dict outputs."
+                    )
```
Amnah199, line 132: Suggestions to improve the error message to provide more meaningful feedback for debugging.
```suggestion
                if not isinstance(outputs, dict):
    raise PipelineRuntimeError(
        f"Component '{component_name}' returned an invalid output type. "
        f"Expected a dict, but got {type(outputs).__name__} instead. "
    )
```

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+                    else:
+                        # No tasks left => schedule anyway (end of pipeline)
+                        # This ensures we don't deadlock forever.
+                        await _schedule_ready_task(component_name)
+                        return
+
+            # -------------------------------------------------
+            # MAIN SCHEDULING LOOP
+            # -------------------------------------------------
+            while True:
```
Amnah199, line 299: If we don't care about partial outputs, we can add a mechanism to cancel all running tasks in case of component fails (or `PipelineRuntimeError` occurs). Something like below:

```
try:
    # Main scheduling loop
except Exception as e:
    for task in running_tasks:
        task.cancel()
    raise e
```

haystack_experimental/core/pipeline/async_pipeline.py
mathislucka: Thanks, this looks good. This part of the code was the same as in Madeesh's original implementation.

haystack_experimental/core/pipeline/async_pipeline.py
mathislucka: I don't think that's necessary. When we raise an Exception, the program will exit anyways and none of the async tasks will be executed anymore.

If the user handles the exception, then it could be confusing if we canceled the async tasks since they will all raise an `asyncio.CancelledError` at the next opportunity. This might obfuscate what's actually happening in the tasks.

haystack_experimental/core/pipeline/async_pipeline.py
```diff
@@ -2,17 +2,592 @@
...
+        pipeline_outputs: Dict[str, Any] = {}
+        running_tasks: Dict[asyncio.Task, str] = {}
+
+        # A set of component names that have been scheduled but not finished:
+        scheduled_components: Set[str] = set()
+
+        # 2) Convert input data
+        prepared_data = self._prepare_component_input_data(data)
+        self._validate_input(prepared_data)
+        inputs_state = self._convert_from_legacy_format(prepared_data)
```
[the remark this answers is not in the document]
mathislucka: Changed it to `_convert_to_internal_format`, does that work?

Amnah199 (approved):
Thanks for the updates @mathislucka . LG!

mathislucka:
@Amnah199 found an issue when running a pipeline concurrently both for the `AsyncPipeline` as well as the sync `Pipeline`. Since component visits are tracked as shared state on the pipeline graph level, concurrent executions might interfere with each other by setting / re-setting visits. I have a fix ready, just adding some tests to make sure it works as intended.

Amnah199:
Cool, thanks for notifying @mathislucka . I can give the PR another quick pass when you push the new fix.

mathislucka:
@Amnah199 problem is fixed.

In both pipeline versions, we now track component visits on a per run basis and don't update the global pipeline state anymore. This is why we also don't need to call `_init_graph` anymore.

The behavioral tests now run twice, once for the sync Pipeline and once for the AsyncPipeline.

I also added two unit tests to test for re-entrance/thread safety for each of the pipeline implementations.

PR #183, opened 2025-01-31
feat: introduce SuperComponent abstractions

mathislucka:
### Related Issues
- split from https://github.com/deepset-ai/haystack-experimental/pull/174

### Proposed Changes:

This PR introduces the scaffolding for SuperComponents. It allows to wrap any pipeline into a friendly component interface and to create your own super components.

### How did you test it?

Unit tests.

### Notes for the reviewer

Ceck out the discussion in the PR [here](https://github.com/deepset-ai/haystack-experimental/pull/174)
### Checklist

- I documented my code


anakin87 (commented):
I've spent some time going through this and the related PRs, but there are a few points I'm still trying to understand before going deeper into the implementation.

- Separation of `SuperComponentBase` and `SuperComponent`
I noticed that `SuperComponentBase` contains most of the functionality (and it's not a component itself, which makes sense). Given that, what's the reasoning behind having `SuperComponent` as a separate class?
(I read through https://github.com/deepset-ai/haystack-experimental/pull/174#issuecomment-2612471571, but I would like to understand better)

- Purpose of `to_super_component_dict`
What is the intended use case for exposing `to_super_component_dict`? Additionally, why do we assume that users would want to manually modify the underlying Pipeline?
From a user's perspective, if I find that `AutoFileConverter` fits my needs, I'd use it as is. If not, I'd likely create a new `SuperComponent` that aligns with my requirements rather than modifying the existing pipeline directly. Curious to hear your thoughts...

mathislucka:
Thanks for the review!

> Separation of SuperComponentBase and SuperComponent
I noticed that SuperComponentBase contains most of the functionality (and it's not a component itself, which makes sense). Given that, what's the reasoning behind having SuperComponent as a separate class?
(I read through https://github.com/deepset-ai/haystack-experimental/pull/174#issuecomment-2612471571, but I would like to understand better)

Users have expressed the need for an abstraction that wraps around pipelines and can be used as a component. Some use cases that this enables:
- organizing your code in large pipelines and encapsulating part of the logic (easier to reason about, reusable, can be tested in isolation)
- using pipelines as tools for an Agent (via `ComponentTool`)

@mikebellerU sums it up quite nicely in [this issue](https://github.com/deepset-ai/haystack/issues/7638):

> Hi -- I have been using haystack to build out some complicated RAG pipelines. They are too complicated to build in a single Pipeline. I would like to be able to "compose" sub-pipelines together. This will allow for building complex pipelines
from smaller ones, and would also allow for reuse of smaller pipelines in various ways.


From an implementation perspective, the `SuperComponent`-class is actually only needed for serialization. It only defines a `to_dict` and `from_dict` method.

> Purpose of to_super_component_dict
What is the intended use case for exposing to_super_component_dict? Additionally, why do we assume that users would want to manually modify the underlying Pipeline?
From a user's perspective, if I find that AutoFileConverter fits my needs, I'd use it as is. If not, I'd likely create a new SuperComponent that aligns with my requirements rather than modifying the existing pipeline directly. Curious to hear your thoughts...

`to_super_component_dict` allows us to add a feature to deepset Studio where users can modify a ready-made super component dynamically in the visual pipeline builder or the yaml editor. We can call the method in the background and then allow the user to modify the underlying pipeline directly.

anakin87:
The problem  you are addressing is quite clear.

I'm still trying to understand if we can improve the design from an OOP perspective.
(I'm not an expert, so a review by @vblagoje could be really helpful here.)

After thinking more about `SuperComponent`, `SuperComponentBase`, and `to_super_component_dict`, I now see a possible alternative that might simplify things:
- Remove `SuperComponent`
- Require implementations of `SuperComponent` to define `to_dict` and `from_dict`. This trades a bit of convenience for clarity.
- Make `to_super_component_dict` and similar deserialization methods internal.

Here's a prototype of the proposed approach. WDYT?

```python
from abc import ABC, abstractmethod
from typing import TypeVar, Dict, Any, Type

T = TypeVar('T', bound='SuperComponentBase')

class SuperComponentBase(ABC):
    def __init__(
            self,
            pipeline: Pipeline,
            input_mapping: Optional[Dict[str, List[str]]] = None,
            output_mapping: Optional[Dict[str, str]] = None,
    ) -> None:
        self.pipeline = pipeline
        self.input_mapping = input_mapping
        self.output_mapping = output_mapping
        # ... rest of initialization ...

    def _to_super_component_dict(self) -> Dict[str, Any]:
        """Internal framework serialization"""
        serialized_pipeline = self.pipeline.to_dict()
        serialized = default_to_dict(
            self, 
            pipeline=serialized_pipeline, 
            input_mapping=self.input_mapping, 
            output_mapping=self.output_mapping
        )
        serialized["type"] = f"{self.__class__.__module__}.{self.__class__.__name__}"
        return serialized

    @classmethod
    def _from_super_component_dict(cls, data: Dict[str, Any]) -> "SuperComponentBase":
        """Internal framework deserialization."""
        # ... deserialization logic ...

    @abstractmethod
    def to_dict(self) -> Dict[str, Any]:
        """Public serialization interface that components must implement."""
        pass

    @classmethod
    @abstractmethod
    def from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
        """Public deserialization interface that components must implement."""
        pass

# Example implementation
class MySuperComponent(SuperComponentBase):
    def __init__(self, pipeline: Pipeline, specific_field: str, **kwargs):
        super().__init__(pipeline, **kwargs)
        self.specific_field = specific_field

    def to_dict(self) -> Dict[str, Any]:
        """Component-specific serialization"""
        return {
            "specific_field": self.specific_field,
            # ... other specific fields ...
        }

    @classmethod
    def from_dict(cls, data: Dict[str, Any]) -> 'MySuperComponent':
        """Component-specific deserialization"""
        return cls(
            specific_field=data["specific_field"],
            # ... other specific fields ...
        )
```

mathislucka:
What is the reason for trying to remove `SuperComponent`?

Your approach would mean that users would have to implement their own `SuperComponent` so that they can have a generic component wrapper around their pipeline. That seems unnecessary because we can just provide that wrapper with a very simple class that defines nothing more than serialization and has a component decorator.

mathislucka:
Maybe also to clarify a bit how I see users mostly using this:

For most use cases I'd see people using the `SuperComponent` directly.
They can just define their pipelines in code and then pass it to a SuperComponent so that they can use it as a building block in other pipelines.
No need to pass any additional configuration parameters to the `SuperComponent` because they can just do all of the configuration on the pipeline-level.


I would only foresee people using `SuperComponentBase` when they create frameworks or libraries (like us). When creating a library, you might want to build components that you can configure from the outside without passing a pipeline to make it more friendly for other developers to use.

Maybe it would make sense to remove `SuperComponentBase` as a top-level import then.

So basically:
- `SuperComponent` -> will be used by 99% of the users
- `SuperComponentBase` -> only used by advanced users creating frameworks or libraries

anakin87:
I now have a better understanding of the requirements:

- a simple out-of-the-box `SuperComponent`: you pass a `Pipeline`, you get an instance of `SuperComponent`
- customization flexibility: users can define specific subclasses of `SuperComponent` (or `SuperComponentBase`) and override behavior as needed.

With this in mind, I still have some uncertainties about the original design. It feels like it might be conflicting with some OOP principles, and it would be great if we can make it easier to understand.

Let me propose another rough idea:
```python
# no SuperComponentBase

@component
class SuperComponent:
    def __init__(self, pipeline: Pipeline, input_mapping: Optional[Dict] = None, ...):
        self.pipeline = pipeline
        self.input_mapping = input_mapping
        # ...

    def _to_super_component_dict(self) -> Dict[str, Any]:
        """Internal framework serialization"""
        return {
            "type": f"{self.__class__.__module__}.{self.__class__.__name__}",
            "pipeline": self.pipeline.to_dict(),
            "input_mapping": self.input_mapping,
            "output_mapping": self.output_mapping
        }

    @classmethod
    def _from_super_component_dict(cls: Type[T], data: Dict[str, Any]) -> T:
        """Internal framework deserialization"""
        return cls(
            pipeline=Pipeline.from_dict(data["pipeline"]),
            input_mapping=data.get("input_mapping"),
            output_mapping=data.get("output_mapping")
        )

    def to_dict(self) -> Dict[str, Any]:
        """Public serialization - can be overridden by advanced users"""
        return self._to_super_component_dict()

    @classmethod
    def from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
        """Public deserialization - can be overridden by advanced users"""
        return cls._from_super_component_dict(data)
```

Do you see any significant drawbacks with this approach?

mathislucka:
Yes, I explored this approach but here is the drawback that I found:

If we (or other users) implement ready-made super components (like `AutoFileProcessor` proposed in the other PR) that inherit from `SuperComponent`, they have to overwrite the `to_dict` and `from_dict` methods as they would otherwise get unexpected results when the component is serialized or deserialized.

If they forget to overwrite the methods, the component would be serialized to a `SuperComponent` and not to the custom component that they actually implemented. We have no other Haystack components that serialize to another class than the original component when `to_dict` is called. That would be very confusing.

Another point is that many components don't have to implement `to_dict` and `from_dict` because the default serde handlers in a pipeline are enough to handle the component. With the proposed change, they would now always have to define these methods. This is inconvenient.

What OOP principle does the base class approach violate?

vblagoje:
What might be throwing one off (us included) here is SuperComponentBase having component word in it but being a plain class having the main task of wrapping/adapting a pipeline and handling things like mapping and serialization.  Then we have component-decorated class SuperComponent, which most users will directly use - but could be named either as is or perhaps it could be renamed to PipelineAdapterComponent. SuperComponentBase could be renamed to PipelineAdapter or something similar.

mathislucka:
Thanks for the feedback.

I did explore the suggested approach already and wasn't happy with the serialization conundrum.

We've clearly outlined the advantages and disadvantages of either approach.

Without `SuperComponentBase`:
+ a single class that can be extended
- users need to define serialization

With `SuperComponentBase`:
+ users don't have to define serialization
- users need to pick the right class for their use case

I'd say that the decision comes down to taste.

I'm happy to implement the single class solution if you think that it's the better approach.

anakin87:
In general, I’m fairly convinced about the single class approach, even though I haven’t tested it extensively, so I’d go with that. If you think it would be helpful, feel free to bring someone else into the discussion.

That said, I don’t want to block progress or spend too much time debating.

mathislucka:
Yeah, let's just go with your approach. It's a small change anyways and since people probably rarely create their own custom super components you might be right and it will be easier in practice with only one class.

haystack_experimental/core/super_component/super_component.py
```diff
@@ -0,0 +1,420 @@
...
+
+        :return: Dictionary containing serialized SuperComponent data
+        """
+        serialized_pipeline = self.pipeline.to_dict()
+        serialized = default_to_dict(
+            self,
+            pipeline=serialized_pipeline,
+            input_mapping=self.input_mapping, output_mapping=self.output_mapping
+        )
+        serialized["type"] = "haystack_experimental.core.super_component.super_component.SuperComponent"
```
mathislucka, line 419: We need to hardcode the type here because dynamic resolution would produce the wrong type when sub-classing this component.

haystack_experimental/core/super_component/super_component.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+class InvalidMappingError(Exception):
+    """Raised when input or output mappings are invalid or type conflicts are found."""
+
+    pass
+
+
+
+class SuperComponent:
```
anakin87, line 20: can we make this a `@component`?

haystack_experimental/core/super_component/super_component.py
anakin87: I am not sure about this aspect. Can you explain it better?

(In any case, we could also fix it later).

haystack_experimental/core/super_component/super_component.py
```diff
@@ -0,0 +1,420 @@
...
+        """
+        Create a pipeline wrapper instance from a dictionary representation.
+
+        :param data: Dictionary containing serialized pipeline wrapper data
+        :return: New PipelineWrapper instance
+        """
+        pipeline = Pipeline.from_dict(data["init_parameters"]["pipeline"])
+        data["init_parameters"]["pipeline"] = pipeline
+
+        return default_from_dict(cls, data)
```
anakin87, line 83: for symmetry with `_to_super_component_dict`, I would also introduce a `_from_super_component_dict` that can be provided as a default implementation of `from_dict`.
I imagine this could be used for internal deserialization (e.g. in Studio). Does this make sense to you?

haystack_experimental/core/super_component/super_component.py
mathislucka: Yes, have a look at [this test](https://github.com/deepset-ai/haystack-experimental/blob/5ce83ca09d31c2ad9a717d3dc405b730de42de57/test/core/super_component/test_super_component.py#L163).

If we resolved the type dynamically like this: `f"{self.__class__.__module__}.{self.__class__.__name__}"` and then subclass the component, calling `_to_super_component_dict` would give us the type of the subclass and not the `SuperComponent`.

haystack_experimental/core/super_component/super_component.py
mathislucka: No, `_from_super_component_dict` would not work. We can not know here how a subclass should be initialized. So _from_super_component_dict would always just yield a plain SuperComponent.

```python
@component
class CustomSuperComponent(SuperComponent):
  def __init__(pipeline, custom_attribute):
    self.custom_attribute = custom_attribute
    super(CustomSuperComponent, self).__init__(pipeline))
  ...

serialized = CustomSuperComponent(pipe, "test")._to_super_component_dict()

deserialized = CustomSuperComponent._from_super_component_dict(serialized)
# This would just yield a normal SuperComponent and not the CustomSuperComponent with the custom_attribute
```

anakin87: I think these points help explain my idea.
- `from_super_component_dict` and `to_super_component_dict` only handle serialization of the core SuperComponent logic (pipeline, mappings)
- custom attributes would be lost, but the core logic is preserved (for this reason, they can be used internally, maybe for wrapping and unwrapping pipelines in Studio)
- users should use `to_dict`/`from_dict` and implement them when they need to handle custom attributes in their custom SuperComponents
- (Since `_from_super_component_dict` is a class method, it returns an instance of whatever class calls it)

I'll provide a better example.

anakin87:
I've iterated a bit on my previous proposal. I think we are almost there.

Does it look reasonable?

```python
from typing import Optional, Dict, List, Any, TypeVar, Type
from haystack import Pipeline, component
from haystack.core.serialization import generate_qualified_class_name
T = TypeVar('T', bound='SuperComponent')

@component
class SuperComponent:
    def __init__(
        self, 
        pipeline: Pipeline,
        input_mapping: Optional[Dict[str, List[str]]] = None,
        output_mapping: Optional[Dict[str, str]] = None,
    ) -> None:
        self.pipeline = pipeline
        self.input_mapping = input_mapping
        self.output_mapping = output_mapping

    def _to_super_component_dict(self) -> Dict[str, Any]:
        """Internal framework serialization - DO NOT OVERRIDE"""
        return {
            "type": generate_qualified_class_name(type(self)),
            "pipeline": self.pipeline.to_dict(),
            "input_mapping": self.input_mapping,
            "output_mapping": self.output_mapping
        }

    @classmethod
    def _from_super_component_dict(cls: Type[T], data: Dict[str, Any]) -> T:
        """Internal framework deserialization - DO NOT OVERRIDE"""
        return cls(
            pipeline=Pipeline.from_dict(data["pipeline"]),
            input_mapping=data.get("input_mapping"),
            output_mapping=data.get("output_mapping")
        )

    def to_dict(self) -> Dict[str, Any]:
        """Public serialization - override if you add custom attributes"""
        return self._to_super_component_dict()

    @classmethod
    def from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
        """Public deserialization - override if you add custom attributes"""
        return cls._from_super_component_dict(data)
    

# Simple SuperComponent
pipeline = Pipeline()
# ...
component = SuperComponent(pipeline)

serialized = component.to_dict()
deserialized = SuperComponent.from_dict(serialized)

internal_serialized = component._to_super_component_dict()
internal_deserialized = SuperComponent._from_super_component_dict(internal_serialized)


# Custom SuperComponent
@component
class CustomSuperComponent(SuperComponent):
    def __init__(self, custom_attribute):
        self.custom_attribute = custom_attribute

        pipeline = Pipeline()
        # ...
        super(CustomSuperComponent, self).__init__(pipeline)

    def to_dict(self) -> Dict[str, Any]:
        # MUST BE IMPLEMENTED
        pass

    @classmethod
    def from_dict(cls, data: Dict[str, Any]):
        # MUST BE IMPLEMENTED
        pass

custom_component = CustomSuperComponent(custom_attribute="test")
serialized = custom_component.to_dict()
deserialized = CustomSuperComponent.from_dict(serialized)

# this is different but preserves the core logic
internal_serialized = custom_component._to_super_component_dict()
internal_deserialized = CustomSuperComponent._from_super_component_dict(internal_serialized)
```

mathislucka:
The example does not make sense for the use case.

The point of `_to_super_component_dict` is that we can get an instance of `SuperComponent` from any component that subclasses it. We do not want `CustomSuperComponent`.

Let me try and make it more clear:
```python
# Simple SuperComponent
pipeline = Pipeline()
# ...
component = SuperComponent(pipeline)

serialized = component.to_dict()
deserialized = SuperComponent.from_dict(serialized)

internal_serialized = component._to_super_component_dict()
internal_deserialized = SuperComponent._from_super_component_dict(internal_serialized)


# Custom SuperComponent
@component
class CustomSuperComponent(SuperComponent):
    def __init__(self, custom_attribute):
        self.custom_attribute = custom_attribute
    
        pipeline = Pipeline()
        # ...
        super(CustomSuperComponent, self).__init__(pipeline)

    def to_dict(self) -> Dict[str, Any]:
        return default_to_dict(self, custom_attribute=custom_attribute)

    @classmethod
    def from_dict(cls, data: Dict[str, Any]):
        return default_from_dict(cls, data)

custom_component = CustomSuperComponent(custom_attribute="test")

# Everything works as expected
serialized = custom_component.to_dict()
deserialized = CustomSuperComponent.from_dict(serialized)

# this has type `CustomSuperComponent`
internal_serialized = custom_component._to_super_component_dict()

# this does not work because we did not serialize the custom attribute
deserialized = CustomSuperComponent.from_dict(internal_serialized)

# this fails too because we try to initialize `CustomSuperComponent()` with a pipeline argument instead custom attribute
internal_deserialized = CustomSuperComponent._from_super_component_dict(internal_serialized)
```

anakin87:
Ok, I see your point now.

haystack_experimental/core/super_component/super_component.py
anakin87: Your comment (https://github.com/deepset-ai/haystack-experimental/pull/183#issuecomment-2640538110) clarified things.

`from_super_component_dict` would be used for internal purposes, so if you think this is not necessary, we can avoid it for the moment.

haystack_experimental/core/super_component/super_component.py
```diff
@@ -0,0 +1,420 @@
...
+        """
+        Warms up the pipeline if it has not been warmed up before.
+        """
+        if not self.warmed_up:
+            self.pipeline.warm_up()
+            self.warmed_up = True
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Convert the pipeline wrapper to a dictionary representation.
```
anakin87, line 66: - let's explain here in docstrings that if you want to create a custom SuperComponent with custom attributes, you MUST appropriately override this method
- nit: let's use super component instead of pipeline wrapper

haystack_experimental/core/super_component/super_component.py
```diff
@@ -0,0 +1,420 @@
...
+
+        :return: Dictionary containing serialized pipeline wrapper data
+        """
+        return self._to_super_component_dict()
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "SuperComponent":
+        """
+        Create a pipeline wrapper instance from a dictionary representation.
```
anakin87, line 75: see above

haystack_experimental/core/super_component/super_component.py
```diff
@@ -0,0 +1,420 @@
...
+
+        :return: Dictionary containing serialized SuperComponent data
+        """
+        serialized_pipeline = self.pipeline.to_dict()
+        serialized = default_to_dict(
+            self,
+            pipeline=serialized_pipeline,
+            input_mapping=self.input_mapping, output_mapping=self.output_mapping
+        )
+        serialized["type"] = "haystack_experimental.core.super_component.super_component.SuperComponent"
```
anakin87, line 419:
```suggestion
        serialized["type"] = generate_qualified_class_name(SuperComponent)
```
let's tie this to the class but dynamically reflect the path

this also needs to imported: `from haystack.core.serialization import generate_qualified_class_name`

mathislucka:
@anakin87 added the dynamic path resolution and the docstring that `to_dict` and `from_dict` need to be overwritten.

The fact that we have to do this in documentation looks like a code smell to me. Therefore, I tried if we could enforce this by making these abstractmethods but `ABCMeta` clashes with `ComponentMeta` so that's not a possibility either.

Anyways, have another look. Hopefully should be ready now.

anakin87 (approved):
I am quite happy with the SuperComponent public API.

The internal implementation details can be refined in the future (and I haven't focused on them much).

Maybe @bilgeyucel wants to review this PR. In any case, from my point of view, it is ready to be merged.

About notebook to showcase this feature (to be done in another PR), I would expect something like
- basic creation of SuperComponent from a Pipeline
- usage examples of ready-made SuperComponents
- advanced: create your own custom SuperComponent (where we also show how to override serde methods)

mathislucka:
I'll merge now to advance this topic. We can make additional changes later when implementing examples.

test/core/super_component/test_super_component.py
```diff
@@ -0,0 +1,192 @@
...
+        input_mapping = {
+            "search_query": ["retriever.query", "prompt_builder.query", "answer_builder.query"]
+        }
+        output_mapping = {
+            "answer_builder.answers": "final_answers"
+        }
+
+        wrapper = SuperComponent(
+            pipeline=rag_pipeline,
+            input_mapping=input_mapping,
```
bilgeyucel, line 87: Why can't I just pass the `data` dict that I would pass to `pipeline.run`? The idea is the same, but from a user's perspective, it's another syntax to learn.

test/core/super_component/test_super_component.py
```diff
@@ -0,0 +1,192 @@
...
+            output_mapping=output_mapping
```
bilgeyucel, line 88: What if I want to access the output of a component that's not the final component of the pipeline? e.g. can I do `"retriever.documents":"source_documents"` in output mapping? Because when I run the pipeline separately, I need to pass `include_outputs_from={"retriever"}` to get that information.

test/core/super_component/test_super_component.py
mathislucka: They can't pass a data dict because this is a component and not a pipeline. The mapping to input/output sockets is necessary so that the users can connect this component to other components.

However, users don't have to provide a mapping, auto-resolution of the mapping should work in most cases.

test/core/super_component/test_super_component.py
mathislucka: I though about this too. We can add an `include_outputs_from` to the __init__ later on.

For now, users would have to build their pipelines in a way where all desired outputs are actually returned from the pipeline.

The idea is to encapsulate the pipeline so that users don't need to care about implementation details of the pipeline anymore (e.g. what components are in it).

PR #203, opened 2025-02-19
chore: add tool related code to prepare Agent PR

mathislucka:
### Proposed Changes:
As per @julian-risch suggestion I'm adding the tool related components and tests to the experimental repository to keep the diff for https://github.com/deepset-ai/haystack-experimental/pull/175 small

### Checklist

- I documented my code


haystack_experimental/components/tools/__init__.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from haystack.components.tools.tool_invoker import ToolInvoker
+
+_all_ = ["ToolInvoker"]
```
sjrl, line 7: I think there should be double underscore
```suggestion
__all__ = ["ToolInvoker"]
```

mathislucka: :D that is wrong in haystack too then. I'll fix it.

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -0,0 +1,242 @@
...
+        for message in messages:
+            tool_calls = message.tool_calls
+            if not tool_calls:
+                continue
+
+            for tool_call in tool_calls:
+                tool_name = tool_call.tool_name
+                tool_arguments = tool_call.arguments
+
+                if not tool_name in self._tools_with_names:
```
sjrl, line 194: small nitpick for pep 8 standard
```suggestion
                if tool_name not in self._tools_with_names:
```

sjrl: Yeah sorry I just realized that a few min ago :D 

Just so I understand why can't we use the ToolInvoker from Haystack directly if this is a straight copy? Is there any need to copy it in haystack-experimental at all?

mathislucka: As discussed offline, there are changes coming to this code with #175 but to keep the diff small we copy over the relevant components first so that only the changes need to be reviewed.

sjrl (approved):
Looks good!

PR #224, opened 2025-03-07
feat: adding a new `Summarizer` component

davidsbatista:
### Proposed Changes:

This PR introduces a new experimental component for text summarization using LLMs:

- Adaptive Chunking: Intelligently splits long documents into optimal chunks based on token limits
- Configurable Detail Level: Allows controlling summary verbosity (0-1 scale)
- Recursive Summarization: Option to use previous summaries as context for better coherence
- The component is designed to handle long documents by breaking them into manageable chunks, summarizing each chunk, and then combining the results - inspired by techniques from the OpenAI cookbook.

### Checklist

- I documented my code


davidsbatista:
I'm using this approach for the [Document Summary Indexing](https://github.com/davidsbatista/haystack-retrieval?tab=readme-ov-file#document-summary-index), and though it might be worth it to have something like this as a component in Haystack - so adding it here for feedback

@bilgeyucel tagging you here as well, as you might find it useful

haystack_experimental/components/summarizer/summarizer.py
```diff
@@ -0,0 +1,409 @@
...
+        if self.generator_api == LLMProvider.OPENAI or self.generator_api == LLMProvider.OPENAI_AZURE:
+            tiktoken_import.check()
+            model_name = self.generator_api_params.get("model", "gpt-4-turbo")
+            encoding = tiktoken.encoding_for_model(model_name)
+            return len(encoding.encode(text))
+
+        # fallback to approximate tokenization - assuming most LLMs average ~4 characters per token
+        return len(text) // 4
+
+    def _chunk_on_delimiter(self, text: str, max_tokens: int, delimiter: str) -> list[str]:
```
sjrl, line 220: We should probably use the recursive splitter here with one delimiter by default right? Or maybe the normal doc splitter.

davidsbatista:
Some more details on the two main features of this summariser and what differentiates it from the `LLMetadataExtractor`

## Chunking strategy

The splitting is a bit more complex, here is a breakdown on how the chunks are created:

The chunking breaks down documents into appropriately sized chunks based on: 
- minimum chunk size
- document length 
- detail level

1. We first calculate how many chunks we would get if we split the document using the minimum chunk size (default: 500 tokens). This is the most granular splitting possible.

```python
max_chunks = len(self._chunk_on_delimiter(text, minimum_chunk_size, chunk_delimiter))
```

2. Then we calculate an actual chunk count base on how much detail we want, `detail` parameter determines how many chunks to actually use:
- At `detail=0`, we use just 1 chunk (the entire document)
- At `detail=1`, we use `max_chunks` (maximum granularity)

```python
min_chunks = 1
num_chunks = int(min_chunks + detail * (max_chunks - min_chunks))
```

3. We then calculate the appropriate chunk size by dividing the document length in tokens by the desired number of chunks
making sure the chunk size is at least the minimum specified size

```python
document_length = self.num_tokens(text)
chunk_size = max(minimum_chunk_size, document_length // num_chunks)
```

Then the document is split into chunks of the calculated size, and each chunk is summarized individually and sequentially.

## Context Preservation

When summarize_recursively is enabled:

- It includes all previous summaries as context when summarizing the next chunk
- Helps maintain continuity and coherence across the entire document
- The LLM can reference information from earlier sections when summarizing later sections
- At the end all summaries are accumulated and eventually joined together

When disabled, each chunk is summarized independently

dfokina:
Let's not forget to add it to pydocs too :)

julian-risch:
I left some comments after briefly going through this PR as I understand you're still discussing if we want to add this component at all. I suggest adding unit tests and integration tests if we decide that we want to add the component before doing another round of reviews.

haystack_experimental/components/summarizer/summarizer.py
medsriha: Agree

haystack_experimental/components/summarizer/summarizer.py
```diff
@@ -0,0 +1,409 @@
...
+
+logger = logging.getLogger(__name__)
+
+SYSTEM_PROMPT = "Rewrite this text in summarized form."
+
+
+class LLMProvider(Enum):
+
+    OPENAI = "openai"
+    OPENAI_AZURE = "openai_azure"
```
Xemnas0, line 33: Throughout Haystack, this has always been referred to "AzureOpenAI", maybe it'd be better to invert it into "AZURE_OPENAI" for consistency

anakin87:
In case we keep working on this component, let's adopt the `ChatGenerator` protocol (https://github.com/deepset-ai/haystack/issues/9058).

haystack_experimental/components/summarizers/summarizer.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        # Calculate document length
+        document_length = self.num_tokens(text)
+
+        # Calculate maximum possible chunks (if we split at minimum_chunk_size)
+        max_chunks = max(1, document_length // minimum_chunk_size)
+        min_chunks = 1
+
+        # Interpolate based on detail parameter
+        num_chunks = int(min_chunks + detail * (max_chunks - min_chunks))
```
julian-risch: Could you please explain a bit more the intuition behind `detail` in this formula in the docstring here and in the docstring of the Summarizer component?

julian-risch: `detail` is used to scale linearly between `min_chunks` and `max_chunks`, correct?

davidsbatista: exactly

davidsbatista:
I see, but my reason for a change was more that we can in the future multiple types of sumarisation techniques and the name should reflect that.

Issue #219, opened 2025-03-04
Investigate how to stream output from Tools


sjrl:
As @mathislucka has worked with Agents he has found that streaming Tool calls to be extremely helpful for transparency and visibility. It allows users to see what is happening while they wait for a final answer and can make it easier to debug by seeing where the Agent may have made an incorrect decision. 

So it would be worthwhile if we could investigate whether we could easily stream the output of tool calls similar as to what we do for our ChatGenerators. 

@mathislucka please add any additional relevant context.

PR #233, opened 2025-03-11
feat: Agent can stream ChatGenerator responses

mathislucka:
### Related Issues

- partially addresses https://github.com/deepset-ai/haystack-experimental/issues/219 but does not add streaming to tools yet

### Proposed Changes:

 Allow users to set a streaming callback on the Agent at init and run time.

### How did you test it?

unit tests

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/agents/agent.py
+++ b/haystack_experimental/components/agents/agent.py
@@ -13,7 +13,7 @@ from haystack.core.component import Component
 from haystack.core.pipeline.base import PipelineError
 from haystack.core.serialization import component_from_dict
 from haystack.dataclasses import ChatMessage
-from haystack.dataclasses.streaming_chunk import StreamingChunk, SyncStreamingCallbackT
+from haystack.dataclasses.streaming_chunk import SyncStreamingCallbackT
 from haystack.utils.callable_serialization import deserialize_callable, serialize_callable
 
 from haystack_experimental.components.tools import ToolInvoker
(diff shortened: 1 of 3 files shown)


haystack_experimental/components/agents/agent.py
```diff
@@ -232,11 +246,17 @@ def _load_component(component_data: Dict[str, Any]) -> Component:
 
         return instance
 
-    def run(self, messages: List[ChatMessage], **kwargs) -> Dict[str, Any]:
+    def run(
+            self,
+            messages: List[ChatMessage],
+            streaming_callback: Optional[SyncStreamingCallbackT] = None,
+            **kwargs
```
sjrl, line 253: small nit on indenting
```suggestion
        self,
        messages: List[ChatMessage],
        streaming_callback: Optional[SyncStreamingCallbackT] = None,
        **kwargs
```

sjrl:
@mathislucka looks good! Only question I have is if the streaming_callback also needs to be passed to the generator init? Or is it passing just at runtime enough?

mathislucka:
I think our expectation generally is that runtime parameters overwrite init parameters so it should be enough. Passing it at init time would be tricky because the chat generator is already initialized. We could do something like `chat_generator.streaming_callback = streaming_callback` but this would rely on all chat generators defining a `streaming_callback` attribute. Once we have a protocol for chat generators we could rethink the approach.

sjrl:
> I think our expectation generally is that runtime parameters overwrite init parameters so it should be enough. 

Yeah that's a good point. I'd say this is good to go then, and the protocol could just be that we require a streaming_callback to be providable at runtime.

sjrl (approved):
Looks good! It would be good to add sede tests when a streaming callback is provided, but that can be tackled later.

PR #235, opened 2025-03-13
feat: Enhance tool error logs and some refactoring

sjrl:
### Related Issues

- fixes #issue-number

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/tools/tool_invoker.py
+++ b/haystack_experimental/components/tools/tool_invoker.py
@@ -187,9 +187,7 @@ class ToolInvoker:
         return ChatMessage.from_tool(tool_result=tool_result_str, error=error, origin=tool_call)
 
     @staticmethod
-    def _inject_state_args(
-        tool: Tool, llm_args: Dict[str, Any], state: State
-    ) -> Dict[str, Any]:
+    def _inject_state_args(tool: Tool, llm_args: Dict[str, Any], state: State) -> Dict[str, Any]:
         """
         Combine LLM-provided arguments (llm_args) with state-based arguments.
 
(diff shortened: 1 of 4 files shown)


mathislucka (approved):
Looks good!

Something we need to pay attention to as we refactor the ToolInvoker more is that it diverges more and more from the same component in Haystack main. I'd be strongly in favor of merging this work into main as soon as possible.

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -191,84 +201,93 @@ def _inject_state_args(
...
-            # If no explicit "message" key, fallback to the entire result
-            return message_content if message_content is not None else result
-        else:
-            # No explicit outputs => each key in result merges into state
-            for k, v in result.items():
-                state.set(k, v)
+        for state_key, config in outputs.items():
+            # Get the source key from the output config, otherwise use the entire result
+            source_key = config.get("source", None)
+            output_value = result if source_key is None else result.get(source_key)
```
sjrl, line 274: Hey @mathislucka I'm not sure about this line. I think it could lead to unintended effects (I'll show below) and it is kind of hard to trigger because we enforce that if `source` is present in `outputs` that it must be a string. 

E.g. I was able to make this test that passes which seems unintuitive (this also works before my changes) 
```python
    def test_merge_tool_outputs_with_output_mapping_2(self):
        weather_tool = Tool(
            name="weather_tool",
            description="Provides weather information for a given location.",
            parameters=weather_parameters,
            function=weather_function,
            outputs={"weather": {}}
        )
        invoker = ToolInvoker(tools=[weather_tool])
        state = State(schema={"weather": {"type": str}})
        merged_results = invoker._merge_tool_outputs(
            tool=weather_tool,
            result={"weather": "sunny", "temperature": 14, "unit": "celsius"},
            state=state
        )
        assert merged_results == {"weather": "sunny", "temperature": 14, "unit": "celsius"}
        assert state.data == {"weather": {"weather": "sunny", "temperature": 14, "unit": "celsius"}}
```

sjrl: We decided that this is expected. Basically if no `source` field is provided in the `outputs` field in the `Tool` then the entire output of the Tool should be saved into that key in State.

Issue #237, opened 2025-03-14
State and Tool proposed Changes from feedback


sjrl:
After discussion with @mathislucka about the State abstraction in the Agent experiment we derived these action items. This list isn't exhaustive, but is meant to completely cover all proposed improvements that are breaking changes so we can bring them in haystack-experimental before moving Agent to Haystack main. 

**Tools** and **ToolInvoker**
- [x] Rename the variables `inputs` and `outputs` to `inputs_from_state` and `outputs_to_state`
- [x] In `outputs_to_state` if you omit `source` then put the entire tool result into that key within State
- [x] drop automatic dict merging
- [x] Keep default extend + append logic for List types defined in State Schema —> fix bug in append logic

**State**
- [x] Don’t dump all results into State by default, only listen to `outputs_to_state` defined in `Tool`
- [x] Change how messages from Tool outputs are handled. We will now store messages in State in a field called `messages` with type `List[ChatMessage]`. This will always have to be there. 
  - [x] Remove specifying a handler for the `messages` field in `outputs_to_state`. For now only allow `source` and we will apply our default handling logic that currently lives in the ToolInvoker. 
  - [x] Later we will allow users to provide a handler.
- [x] Once State includes `messages` we can probably simplify agent e.g. one less Branch Joiner
- [x] Assign default handlers at State initialization. If a user were to print the State schema it would be ideal for them to be able to see and inspect the default handlers. Currently default behavior is in ToolInvoker and not transparent.

PR #239, opened 2025-03-17
feat: Updates to insertion of values into `State`

sjrl:
### Related Issues

- partially addresses https://github.com/deepset-ai/haystack-experimental/issues/237

### Proposed Changes:

 
Addresses these points in the check list

**Tools** and **ToolInvoker**
- [x] In `outputs_to_state` if you omit `source` then put the entire tool result into that key within State
- [x] drop automatic dict merging
- [x] Keep default extend + append logic for List types defined in State Schema —> fix bug in append logic

**State**
- [x] Don’t dump all results into State by default, only listen to `outputs_to_state` defined in `Tool`
- [x] Assign default handlers at State initialization. If a user were to print the State schema it would be ideal for them to be able to see and inspect the default handlers. Currently default behavior is in ToolInvoker and not transparent.

### How did you test it?

Updated tests

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/tools/tool_invoker.py
+++ b/haystack_experimental/components/tools/tool_invoker.py
@@ -247,7 +247,7 @@ class ToolInvoker:
         if not isinstance(result, dict):
             return result
 
-        # If there is no specific `outputs` mapping, we jus
+        # If there is no specific `outputs` mapping, we just return the full result
         if not hasattr(tool, "outputs") or not isinstance(tool.outputs, dict):
             return result
 
--- a/haystack_experimental/components/tools/tool_invoker.py
+++ b/haystack_experimental/components/tools/tool_invoker.py
@@ -227,9 +227,9 @@ class ToolInvoker:
         It also determines what message, if any, should be returned for further processing in a conversation.
 
         Processing Steps:
-        1. If `result` is not a dictionary, it is treated as a single output and returned directly.
-        2. If the `tool` does not define an `outputs` mapping, the entire `result` dictionary is merged into the state
-           as key-value pairs. The return value in this case is simply the full `result` dictionary.
+        1. If `result` is not a dictionary, nothing is stored into state and the full `result` is returned.
+        2. If the `tool` does not define an `outputs` mapping nothing is stored into state.
+           The return value in this case is simply the full `result` dictionary.
         3. If the tool defines an `outputs` mapping (a dictionary describing how the tool's output should be processed),
            the method delegates to `_handle_tool_outputs` to process the output accordingly.
            This allows certain fields in `result` to be mapped explicitly to state fields or formatted using custom
@@ -247,10 +247,8 @@ class ToolInvoker:
         if not isinstance(result, dict):
             return result
 
-        # If there is no specific `outputs` mapping, we default to merging all result keys into the state.
+        # If there is no specific `outputs` mapping, we jus
         if not hasattr(tool, "outputs") or not isinstance(tool.outputs, dict):
-            for key, value in result.items():
-                state.set(key, value)
             return result
 
         # Handle tool outputs with specific mapping for message and state updates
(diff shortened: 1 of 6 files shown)


haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -227,9 +227,9 @@ def _merge_tool_outputs(self, tool: Tool, result: Any, state: State) -> Any:
         It also determines what message, if any, should be returned for further processing in a conversation.
 
         Processing Steps:
-        1. If `result` is not a dictionary, it is treated as a single output and returned directly.
-        2. If the `tool` does not define an `outputs` mapping, the entire `result` dictionary is merged into the state
-           as key-value pairs. The return value in this case is simply the full `result` dictionary.
+        1. If `result` is not a dictionary, nothing is stored into state and the full `result` is returned.
+        2. If the `tool` does not define an `outputs` mapping nothing is stored into state.
```
anakin87, line 231: `outputs_to_state`, no?

haystack_experimental/components/tools/tool_invoker.py
```diff
@@ -247,10 +247,8 @@ def _merge_tool_outputs(self, tool: Tool, result: Any, state: State) -> Any:
         if not isinstance(result, dict):
             return result
 
-        # If there is no specific `outputs_to_state` mapping, we default to merging all result keys into the state.
+        # If there is no specific `outputs` mapping, we just return the full result
```
anakin87, line 250:
```suggestion
        # If there is no specific `outputs_to_state` mapping, we just return the full result
```

haystack_experimental/dataclasses/state.py
```diff
@@ -110,50 +130,32 @@ def set(
...
-        if force or (not declared_type and not declared_handler and not handler_override):
-            self._data[key] = value
-            return
+        # If key not in schema, we throw an error
+        definition = self.schema.get(key, None)
+        if definition is None:
+            logger.error(
+                "State: Key '{state_key}' not found in schema. Schema: {schema}", state_key=key, schema=self.schema
+            )
+            raise ValueError(f"State: Key '{key}' not found in schema. Schema: {self.schema}")
```
anakin87, line 153: if we want to raise an exception, logging the same message seems superfluous

anakin87 (commented):
I only left some minor comments

haystack_experimental/components/tools/tool_invoker.py
sjrl: ahh yes good catch thanks!

haystack_experimental/dataclasses/state.py
sjrl: Let me check with @wochinge on that. I was under the impression that it could still be useful to log the error even though we are raising.

sjrl: Sounds good! Will remove

anakin87 (approved):
Looks good.

I would appreciate it if @mathislucka could also take a look.

mpangrazzi (approved):
Looks good to me as well!

haystack_experimental/dataclasses/state.py
```diff
@@ -110,50 +127,29 @@ def set(
...
-
-        if force or (not declared_type and not declared_handler and not handler_override):
-            self._data[key] = value
-            return
+        # If key not in schema, we throw an error
+        definition = self.schema.get(key, None)
+        if definition is None:
+            raise ValueError(f"State: Key '{key}' not found in schema. Schema: {self.schema}")
 
+        # if the current value is None just replace
```
mathislucka, line 149: Is this comment accurate? Aren't we getting the current value from state?

mathislucka (approved):
Looks good.

Only one minor comment.

haystack_experimental/dataclasses/state.py
sjrl: good catch! this was a leftover

PR #244, opened 2025-03-19
refactor: Small updates. Make pipeline internal, add check for warm_up

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 
- Make underlying pipeline internal. In case we want to change the logic in the future we can do so more easily without breaking changes.
- Add missing check for running warm_up in the run method
- Use thread safe import in the load_component function

### How did you test it?

Existing tests

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/agents/agent.py
+++ b/haystack_experimental/components/agents/agent.py
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import importlib
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from haystack import Pipeline, component, default_from_dict, default_to_dict, logging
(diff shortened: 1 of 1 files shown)


haystack_experimental/components/agents/agent.py
```diff
@@ -262,6 +261,9 @@ def run(
         :param kwargs: Additional keyword arguments matching the defined input types
         :return: Dictionary containing messages and outputs matching the defined output types
         """
+        if not self._is_warmed_up:
+            raise RuntimeError("The component Agent wasn't warmed up. Run 'warm_up()' before calling 'run()'.")
```
sjrl, line 265: @anakin87 what's your take on this? It's not technically needed since we eventually call self._pipeline.run() which does call `warm_up` internally. I guess good to have in case we ever decide to not use `Pipeline` to run Agent?

anakin87: It seems consistent with other components in our codebase, so I'm OK with this approach.

anakin87 (approved):
👍

haystack_experimental/components/agents/agent.py
sjrl: After talking with @julian-risch for a similar topic about SuperComponents that it doesn't make sense to have the check since it technically isn't needed since pipeline.run will warm up the components. 

So removing for now, but could come back if we go with my proposal [here](https://github.com/deepset-ai/haystack-experimental/pull/245)

PR #245, opened 2025-03-19
Proposal for changing internal working of Agent

sjrl:
### Related Issues

@anakin87 and I had a discussion about the internals of the Agent and that it was difficult to understand the logic of the underlying pipeline. 

Additionally, the mermaid picture of the pipeline powering the Agent seems to me to be similarly confusing
test

### Proposed Changes:

 
 
This my proposal for changing the Agent to use an explicit python loop to run the agent rather than using a pipeline.

Pros of this approach:
- Makes the logic easier to follow and understand. I also believe that this will make it easier for us to extend the functionality of the Agent in the future and allow for more sophisticated error handling (if we so desire)
- Reduce the number of needed Haystack components from 5 to 2
- No longer have to understand or extend Jinja2 templates to check exit conditions
- Can explicitly use `max_runs_per_component` (could rename to `max_loops`) to set how many loops we allow the Agent to run before forcing an exit. I think calling this `max_loops` is more understandable and now we explicitly use it to set the number of loops rather than relying on hitting an implicit exit which depends on how many times one of the components in the pipeline is run.  

Cons of this approach:
- We lose builtin tracing by not using Pipeline. However, we now only have two Haystack components being used in the for loop instead of 5 so I would think traces would be easier to understand once we add support for it. 
- For the upcoming pipeline breakpoints feature we'd need to manually implement that as well for Agent if we'd like to add breakpoints within the loop. For this though I think there are natural places to have a breakpoint which would be where the exit conditions are checked, so one place would be after the chat generator and the other right after the tool_invoker. And at both of these places we would have all information needed to restart the generation.  

### How did you test it?

All existing tests pass

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/agents/agent.py
+++ b/haystack_experimental/components/agents/agent.py
@@ -2,19 +2,17 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import importlib
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
-from haystack import Pipeline, component, default_from_dict, default_to_dict, logging
+from haystack import component, default_from_dict, default_to_dict, logging
 from haystack.components.generators.chat.openai import OpenAIChatGenerator
-from haystack.components.joiners import BranchJoiner
-from haystack.components.routers.conditional_router import ConditionalRouter
 from haystack.core.component import Component
 from haystack.core.pipeline.base import PipelineError
 from haystack.core.serialization import component_from_dict
 from haystack.dataclasses import ChatMessage
 from haystack.dataclasses.streaming_chunk import SyncStreamingCallbackT
 from haystack.utils.callable_serialization import deserialize_callable, serialize_callable
+from haystack.utils import type_serialization
 
 from haystack_experimental.components.tools import ToolInvoker
 from haystack_experimental.dataclasses.state import State, _schema_from_dict, _schema_to_dict, _validate_schema
@@ -58,8 +56,9 @@ class Agent:
     ```
     """
 
-    def __init__(  # pylint: disable=too-many-positional-arguments
+    def __init__(
         self,
+        *,
         chat_generator: Union[OpenAIChatGenerator, "AnthropicChatGenerator"],
         tools: Optional[List[Tool]] = None,
         system_prompt: Optional[str] = None,
@@ -106,7 +105,7 @@ class Agent:
             output_types[param] = config["type"]
         component.set_output_types(self, **output_types)
 
-        self._initialize_pipeline()
+        self.tool_invoker = ToolInvoker(tools=self.tools, raise_on_failure=self.raise_on_tool_invocation_failure)
 
         self._is_warmed_up = False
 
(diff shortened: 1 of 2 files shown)


anakin87 (commented):
I left some minor comments on the implementation, but overall I really like how this proposal makes it easier to understand how the Agent works and simplifies maintenance. I think this is a good direction.

haystack_experimental/components/agents/agent.py
```diff
@@ -273,13 +214,38 @@ def run(
...
+                    and not tool_messages[0].tool_call_result.error
+                )
+                if should_exit:
+                    return {"messages": messages + llm_messages + tool_messages, **state.data}
+
+            # 5. Combine messages, llm_messages and tool_messages and send to the ChatGenerator
+            messages = messages + llm_messages + tool_messages
+            counter += 1
+
+        logger.warning(f"Agent exceeded maximum runs per component ({self.max_runs_per_component}), stopping.")
```
mpangrazzi, line 250: Question: why not raising an exception here (after exceeding `self.max_runs_per_component`)?

Maybe it would be nice to add a test for multiple subsequent tool invocation failures.

mpangrazzi (commented):
I like a lot the simpler logic! I've left a comment

haystack_experimental/components/agents/agent.py
sjrl: Good question, unsure if we should raise an exception or if we exit with a warning message. 

For example, I could see a user just want to continue the generation by calling the agent again with the output returned here. And just because we hit the max runs per component doesn't mean it's an error per say.

sjrl: > Maybe it would be nice to add a test for multiple subsequent tool invocation failures.

I totally agree, if we are all in support of going in this direction, I'll spend the time to expand the tests :)

anakin87 (approved):
I fully agree with this direction. Feel free to collect other opinions.

haystack_experimental/components/agents/agent.py
mpangrazzi: @sjrl 

> For example, I could see a user just want to continue the generation by calling the agent again with the output returned here. And just because we hit the max runs per component doesn't mean it's an error per say.

Makes sense! What do you think about adding an additional option for raising here then?

mpangrazzi (approved):

haystack_experimental/components/agents/agent.py
sjrl: Yeah I think that could make sense. To clarify, we think this could be helpful for debugging purposes? So if I have an Agent in a Pipeline that we stop the execution in case this happens instead of continuing?

mpangrazzi: yes, IMHO it could be useful for tool debugging purposes. Example: you have a tool which is not giving a good-enough answer to a LLM or so and it will be called repeatedly (and you don't want that). But ofc one could also simply gracefully answer after N tool calls failures. Your choice, nothing strictly needed ;)

vblagoje (commented):
Yes good direction but, for me at least, not only for simplicity but for users to see how they could make this component custom, react, tree of thought, whatever agent du jour is - it can be added easily.

haystack_experimental/components/agents/agent.py
```diff
@@ -273,13 +214,38 @@ def run(
...
+            # TODO Possible for LLM to return multiple messages (e.g. multiple tool calls)
+            #      So exit condition could be missed if it's not the first message
+            if self.exit_condition != "text" and (
+                llm_messages[0].tool_call.tool_name == self.exit_condition
+                and not tool_messages[0].tool_call_result.error
+            ):
+                return {"messages": messages + llm_messages + tool_messages, **state.data}
+
+            # 5. Combine messages, llm_messages and tool_messages and send to the ChatGenerator
+            messages = messages + llm_messages + tool_messages
```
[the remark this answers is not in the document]
sjrl: Great idea! I'm actually already working on this in a separate PR. Something Mathis and I already discussed based on feedback from the community. We will be moving the messages to State and allowing users to pass a handler to the State schema to customize this processing. We will set the default to be merging as we have now.

tstadel:
I generally like simplifications, however with this change we lose any ability for customization. We just need to be aware of that. To be honest, I don't know if customization is actually needed for agents. I just see, that before, you could simply change the underlying pipeline by accessing it after creation of the Agent. This left way more options than the ones given by init params (e.g. also subclassing the Agent class and overwriting/changing the pipeline). Additionally, the wrapped pipeline left a door open to convert this into a proper SuperComponent at some point. I don't want to be blocking here, but please keep in mind, that such a requirement (being able adjust the Agent's inner workings beyond init params and/or to convert it into a SuperComponent) might not too far from now.

sjrl:
@tstadel I don't fully agree. I think it's usually bad practice to change the behavior of something by changing a class variable that's not exposed via the init. Additionally, I don't see how we could make an Agent just a SuperComponent when we have something like State needed for it to work. 

Personally I think this makes for easier customization since users can easily make a custom component agent that builds off the python code here. The logic is much easier to understand which as @vblagoje points out would make it straightforward to extend. If we limit users to have to using a Pipeline to get the same behavior I think we'd be adding an unnecessary blocker.

tstadel:
> @tstadel I don't fully agree. I think it's usually bad practice to change the behavior of something by changing a class variable that's not exposed via the init. Additionally, I don't see how we could make an Agent just a SuperComponent when we have something like State needed for it to work. 
> 
> Personally I think this makes for easier customization since users can easily make a custom component agent that builds off the python code here. The logic is much easier to understand which as @vblagoje points out would make it straightforward to extend. If we limit users to have to using a Pipeline to get the same behavior I think we'd be adding an unnecessary blocker and more confusion. The current implementation heavily relies on understanding how Jinja2 templates work which have historically confused by Haystack and dC users. 

I see your points, especially Jinja2 templating as big hurdle I can fully understand for most users. But this approach would send users straight to code for any customization. Isn't that the opposite direction as you need to know even more about coding? 
Additionally there is no shortcut anymore (this is what I would call any Jinja2 usages currently) e.g. for dC users. Jinja2 shortcuts are far from ideal, I agree as you need to understand Jinja2 coding and there's little support for writing them at this time. But it has proven really useful for expert users. And that's where I see the problem with this approach: It'd block non-code users completely for customization. This would be a design decision as for e.g. pipeline.run logic.
I wonder if there's no way to strip any Jinja2 usage from the agents pipeline instead. 
I don't want to sound opinionated here, but I clearly see this as drawback for platform users in particular. Reversing this decision might be possible but hard as mapping back the now developing code to pipeline in the future might be difficult.

tstadel:
Sorry I forgot to comment on state vs supercomponent: How about making state a citizen of pipeline?

ju-gu:
I would for simplicity favor Sebs approach. I wonder how much customization we need in agents. From my perspective there are different agent architectures, we can build components for. I would guess that only in rare cases users would want to change those, as it is now with other components as well. From my experience non and little-code users won't be able to change the agent architecture unrelated whether we use pipelines or code. They are usually happy when there are out of the box offers and the level of customization you have with choosing tools. 

I think adding logs/tracing is a must though because you really want to know whats going on inside. Probably now also easier to debug as I don't know what happens if an error is inside the pipeline of a component.

julian-risch (approved):
The simplicity of the current implementation looks great! Keep in mind that adding support for tracing and checkpointing within the Agent component, which I see as a must, it will get a bit more complex.
Other, more complex kinds of agents (react, tree of thought) can be built as custom components or we add them as ready-made components later following the implementation of this proposal. I understand that no-code customization is quite limited then. 
I think a lot of feedback has already been addressed and I like that! In particular, allowing customized processing of the aggregation of messages + llm_messages + tool_messages.

PR #260, opened 2025-04-02
fix: starting from clean state

davidsbatista:
- new strategy, remove `sender` from serialisation
- start new deserialisation strategy

### DIFF
--- a/haystack_experimental/core/pipeline/pipeline.py
+++ b/haystack_experimental/core/pipeline/pipeline.py
@@ -410,48 +410,6 @@ class Pipeline(PipelineBase):
 
         return value
 
-
-    @staticmethod
-    def remove_sender(data):
-        """
-        Removes the 'sender' key from the data.
-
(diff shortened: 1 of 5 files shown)


haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -450,70 +474,6 @@ def _deserialize_component_input(value):  # noqa: PLR0911
...
-                        return [find_serialized_objects(item) for item in lst]
-                    return lst
-
-                result = find_serialized_objects(value)
-
-                while isinstance(result, list) and len(result) == 1 and isinstance(result[0], list):
-                    result = result[0]
-                return result
-
-            return [Pipeline._deserialize_component_input(item) for item in value]
```
Amnah199, line 505: I think this was working fine for deserializing haystack dataclasses. If you want to remove it for now, its okay. We can reintroduce it with the refactored solution.

davidsbatista: yes

PR #266, opened 2025-04-03
Fix major bugs in serialization

Amnah199:
Chat messages are not serialized properly.

### DIFF
--- a/haystack_experimental/core/pipeline/pipeline.py
+++ b/haystack_experimental/core/pipeline/pipeline.py
@@ -79,7 +79,8 @@ class Pipeline(PipelineBase):
 
         # add component_inputs to inputs
         breakpoint_inputs = inputs.copy()
-        breakpoint_inputs[component_name] = self._serialize_component_input(component_inputs, instance)
+        # breakpoint_inputs[component_name] = self._serialize_component_input(component_inputs, instance)
+        breakpoint_inputs[component_name] = component_inputs
         if breakpoints and not self.resume_state:
             self._check_breakpoints(breakpoints, component_name, instance, component_visits, breakpoint_inputs)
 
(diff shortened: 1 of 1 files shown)


haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -433,7 +433,7 @@ def transform_json_structure(data, component):
         elif isinstance(data, list):
             # transform each item in the list
             transformed = [Pipeline.transform_json_structure(item, component) for item in data]
-
+            return transformed
             # NOTE: if the list has only one item, return just that item unless we are dealing with an AnswerBuilder
             if len(transformed) == 1:
                 # NOTE: This is a workaround only for the AnswerBuilder component
```
Amnah199, line 439: This part is whats causing the issue with deserializing ChatMessages. The `_content` is supposed to be a list but this step removes the list around it. I am checking how to fix this without causing other components to crash.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -516,6 +516,7 @@ def _deserialize_component_input(value):  # noqa: PLR0911
         if isinstance(value, dict):
             if "_type" in value and "_module" in value:
                 type_name = value.pop("_type")
+                value.pop("_module")
```
Amnah199, line 519: I think we are not using _module anywhere so we can perhaps not store it.

haystack_experimental/core/pipeline/pipeline.py
davidsbatista: ah, good catch! I think it's enough: 

```python
if isinstance(component, AnswerBuilder) or isinstance(component, ChatMessage):
```

Amnah199: I guess more instances will keep coming up where removing a list will raise an error. This hack will also keep getting longer. WDYT?

Amnah199: Ideally the original type of value should be preserved, whether a str, int or list.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -79,7 +79,8 @@ def _run_component(  # pylint: disable=too-many-positional-arguments
 
         # add component_inputs to inputs
         breakpoint_inputs = inputs.copy()
-        breakpoint_inputs[component_name] = self._serialize_component_input(component_inputs, instance)
+        # breakpoint_inputs[component_name] = self._serialize_component_input(component_inputs, instance)
```
davidsbatista, line 82: this line can be removed

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -421,29 +422,28 @@ def transform_json_structure(data, component):
...
+            # If the original list has exactly one element and that element was a dict
+            # with 'sender' and 'value', then unwrap the list.
 
-            # NOTE: if the list has only one item, return just that item unless we are dealing with an AnswerBuilder
-            if len(transformed) == 1:
-                # NOTE: This is a workaround only for the AnswerBuilder component
-                # we need to wrap the value in a list
-                if isinstance(component, AnswerBuilder):
-                    return transformed
+            if len(data) == 1 and isinstance(data[0], dict) and "value" in data[0] and "sender" in data[0]:
```
davidsbatista, line 441: I think this line + the return can run before the line above

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -404,15 +404,13 @@ def _remove_unserialisable_data(value):
...
                 value.meta["usage"].pop("prompt_tokens_details", None)
 
-        if isinstance(value, dict) and "type" in value:  # noqa: SIM102
-            if value["type"] == "haystack.dataclasses.answer.GeneratedAnswer":  # noqa: SIM102
-                if "meta" in value and "usage" in value["meta"]:    # noqa: SIM102
-                    value["meta"]["usage"].pop("completion_tokens_details", None)
-                    value["meta"]["usage"].pop("prompt_tokens_details", None)
+        if isinstance(value, GeneratedAnswer):  # noqa: SIM102
+            if value.meta and "usage" in value.meta:  # noqa: SIM102
+                value.meta.pop("usage", None)
```
Amnah199, line 413: As we are not calling serilization on component inputs, this needed to be changed for AnswerJpiner tests to pass.

davidsbatista (approved):

PR #269, opened 2025-04-07
feat: adding `Pipeline` checkpoints

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/8972
### Proposed Changes: Pipeline Debugging and State Management Enhancements

This PR introduces improvements to the Pipeline, adding debugging capabilities and state management. The changes make it easier to debug pipelines.

- It adds the ability to set breakpoints at specific component visits, but the pipeline execution is stopped at the defined breakpoints.
- A breakpoint generates a state, a JSON file with the current pipeline state at the specified breakpoint. A Pipeline can then be resumed from this state file

## Usage Examples

```python
# Setting breakpoints
pipeline.run(
    data={"input": "value"},
    breakpoints={("component_name", 0)},  # Break at the first visit
    debug_path="debug_states/"
)
```

This will generate a JSON with the complete pipeline state before the next component is run, i.e.: the one receiving the output of the component set in the breakpoint

```python
# Resuming from a saved state
state = Pipeline.load_state("debug_states/component_state.json")
pipeline.run(
    data={"input": "value"},
    resume_state=state
)
```

### How did you test it?

We currently test this approach with six different pipelines, and for each component in each pipeline i.e.: the pipeline must stop and resume at each component
  - Hybrid RAG Pipeline - uses `DocumentJoiner`
  - Pipelines that use other joiners: `BranchJoiner`, `AnswerJoiner`, `ListJoiner`, `StringJoiner`
  - A Pipeline with a loop

### Notes:

- I'm opening this PR against another branch that only has the current `pipeline.py` from haystack main - this is so that the reviewers can better see what's being added to the original/current `pipeline.py` instead of just seeing a whole new file.
- There are still two edge cases for the `DocumentJoiner` and the `BranchJoiner` where an extra list is added during the serialisation/deserialisation process - we are still investigating/fixing those cases.
- The macOS tests fail with a `Fatal Python error: Segmentation fault` - also investigating/fixing 

### Checklist

- I documented my code


davidsbatista:
@anakin87 this is a big one, I know, we can go on a call if you wish, and I can guide you through the code changes and demo what's done.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +303,382 @@ def run(  # noqa: PLR0915, PLR0912
...
+            if len(data) == 1 and isinstance(data[0], dict) and "value" in data[0] and "sender" in data[0]:
+                return transformed[0]
+            return transformed
+
+        else:
+            # For other data types, just return the value as is.
+            return data
+
+    @staticmethod
+    def _serialize_component_input(value, component):
```
julian-risch, line 458: Let's use type annotations for the input params and returns here too. They make it easier to understand the code.

haystack_experimental/core/errors.py
```diff
@@ -0,0 +1,59 @@
...
+
+class DeserializationError(Exception):
+    pass
+
+
+class SerializationError(Exception):
+    pass
+
+
+class PipelineBreakException(Exception):
```
julian-risch, line 50: `PipelineBreakException` with just `Break` in the name doesn't explain the cause very well. We should decide on how to consistently use Breakpoint, Checkpoint, State as terms.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -175,6 +219,16 @@ def run(  # noqa: PLR0915, PLR0912
...
             output is included.
+
+        :param breakpoints:
+            Set of tuples of component names and visit counts at which the pipeline should break execution.
+
+        :param resume_state:
+            A dictionary containing the state of a previously saved pipeline execution.
+
+        :param debug_path:
+            Path to the directory where the pipeline state should be saved.
```
anakin87, line 230: I need to understand better this point. We can discuss this in the call.
Why is the pipeline responsible for saving the pipeline state to disk?
Have we considered to simply return a dict instead?

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -99,7 +138,12 @@ def _run_component(
             return cast(Dict[Any, Any], component_output)
 
     def run(  # noqa: PLR0915, PLR0912
-        self, data: Dict[str, Any], include_outputs_from: Optional[Set[str]] = None
+        self,
+        data: Dict[str, Any],
+        include_outputs_from: Optional[Set[str]] = None,
+        breakpoints: Optional[Set[Tuple[str, Optional[int]]]] = None,
+        resume_state: Optional[Dict[str, Any]] = None,
```
julian-risch, line 145: if `breakpoints` and `resume_state` are both set, let's show at least a warning that `breakpoints` will be ignored (if that's the behavior)

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +303,382 @@ def run(  # noqa: PLR0915, PLR0912
...
+            logger.info(f"Successfully loaded pipeline state from: {file_path}")
+            return state
+        except FileNotFoundError:
+            raise FileNotFoundError(f"File not found: {file_path}")
+        except json.JSONDecodeError as e:
+            raise json.JSONDecodeError(f"Invalid JSON file {file_path}: {str(e)}", e.doc, e.pos)
+        except IOError as e:
+            raise IOError(f"Error reading {file_path}: {str(e)}")
+
+    def _validate_components_state(self, resume_state: Dict[str, Any]) -> None:
```
julian-risch, line 609: _validate_pipeline_state instead?

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -45,10 +57,37 @@ def _run_component(
...
 
+        # Deserialize the inputs if they are passed in resume state
+        # this check will prevent other inputs generated at runtime from being deserialized
+        if self.resume_state and component_name in self.resume_state["pipeline_state"]["inputs"].keys():
+            for key, value in component_inputs.items():
+                component_inputs[key] = Pipeline._deserialize_component_input(value)
+
+        # add component_inputs to inputs
+        breakpoint_inputs = inputs.copy()
+        breakpoint_inputs[component_name] = Pipeline._remove_unserialisable_data(component_inputs)
```
julian-risch, line 87: We're mixing serialize / serialise. For consistency, we should use _remove_unserializable_data

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +303,382 @@ def run(  # noqa: PLR0915, PLR0912
...
+        :param breakpoints: Set of tuples of component names and visit counts at which the pipeline should stop.
+        :param component_name: Name of the component to check.
+        :param component_visits: The number of times the component has been visited.
+        :param inputs: The inputs to the pipeline.
+        :raises PipelineBreakException: When a breakpoint is triggered, with component state information.
+        """
+        matching_breakpoints = [bp for bp in breakpoints if bp[0] == component_name]
+        for bp in matching_breakpoints:
+            visit_count = bp[1]
+            # break at every visit if visit_count is -1
```
julian-risch, line 398: Let's remove this feature because we will raise an Exception when we break for the first time. Users can stop at every visit by resuming the pipeline run and increasing the `visit_count` in `breakpoints={("component_name", visit_count)},`

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +303,382 @@ def run(  # noqa: PLR0915, PLR0912
...
-                    pipeline_outputs[component_name] = deepcopy(component_pipeline_outputs)
-                if self._is_queue_stale(priority_queue):
-                    priority_queue = self._fill_queue(ordered_component_names, inputs, component_visits)
+    def _validate_breakpoints(self, breakpoints: Set[Tuple[str, Optional[int]]]) -> Set[Tuple[str, int]]:
+        """
+        Validates the breakpoints passed to the pipeline.
 
-            return pipeline_outputs
+        Make sure they are all valid components registered in the pipeline,
+        If the visit is not given, it is assumed to be 0, it will break on the first visit.
```
julian-risch, line 366: we should document that also in the docstring of the pipeline input params.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -189,32 +243,52 @@ def run(  # noqa: PLR0915, PLR0912
...
 
-        # We need to access a component's receivers multiple times during a pipeline run.
-        # We store them here for easy access.
-        cached_receivers = {name: self._find_receivers_from(name) for name in ordered_component_names}
+        else:
+            self._validate_components_state(self.resume_state)
+            data = self._prepare_component_input_data(self.resume_state["pipeline_state"]["inputs"])
+            component_visits = self.resume_state["pipeline_state"]["component_visits"]
+            self.ordered_component_names = self.resume_state["pipeline_state"]["ordered_component_names"]
+            cached_receivers = {name: self._find_receivers_from(name) for name in self.ordered_component_names}
```
julian-risch, line 286: We're doing this in the `if` case and in `else` case. If we move the log message a few lines up, we can move 
`cached_receivers = {..` out of the if / else.

haystack_experimental/core/pipeline/pipeline.py
anakin87: I would like if @tstadel could confirm this requirement - writing JSON to disk.

In case it is necessary, I would make the interface a bit more consistent between writing and reading.
Something like:

```python
pipeline.run(
    data={"input": "value"},
    resume_state="my_state.json" # no need to call Pipeline.load_state
)
```

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +303,382 @@ def run(  # noqa: PLR0915, PLR0912
...
+        from datetime import datetime
+
+        if isinstance(self.debug_path, str):
+            self.debug_path = Path(self.debug_path)
+        if not isinstance(self.debug_path, Path):
+            raise ValueError("Debug path must be a string or a Path object.")
+        self.debug_path.mkdir(exist_ok=True)
+
+        dt = datetime.now()
+        file_name = Path(f"{component_name}_state_{dt.strftime('%Y_%m_%d_%H_%M_%S')}.json")
```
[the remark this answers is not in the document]
Amnah199: I like the second option better. @davidsbatista what do you think?

haystack_experimental/core/errors.py
Amnah199: Lets stick to `Breakpoint`.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +301,382 @@ def run(  # noqa: PLR0915, PLR0912
...
+                value.meta["usage"].pop("prompt_tokens_details", None)
+
+        if isinstance(value, GeneratedAnswer):  # noqa: SIM102
+            if value.meta and "usage" in value.meta:  # noqa: SIM102
+                value.meta.pop("usage", None)
+
+        return value
+
+    @staticmethod
+    def transform_json_structure(data: Any, component: Any) -> Any:
```
Amnah199, line 428: @davidsbatista this `component` param is not actually being used in the method. Should we remove it or this was added for a specific reason?

Amnah199: I think I'll remove it from all the methods including `save_state`.

haystack_experimental/core/pipeline/pipeline.py
Amnah199: I added the `load_state` call in the resuming logic so the user can just pass the json file path as you mentioned.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -45,10 +57,37 @@ def _run_component(
...
+        # this check will prevent other inputs generated at runtime from being deserialized
+        if self.resume_state and component_name in self.resume_state["pipeline_state"]["inputs"].keys():
+            for key, value in component_inputs.items():
+                component_inputs[key] = Pipeline._deserialize_component_input(value)
+
+        # add component_inputs to inputs
+        breakpoint_inputs = inputs.copy()
+        breakpoint_inputs[component_name] = Pipeline._remove_unserialisable_data(component_inputs)
+        if breakpoints and not self.resume_state:
+            self._check_breakpoints(breakpoints, component_name, instance, component_visits, breakpoint_inputs)
```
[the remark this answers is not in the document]
Amnah199: For now I removed passing component instance as param, as we are not really using it in any method.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -45,10 +57,37 @@ def _run_component(
             component_name=component_name, component=component, inputs=inputs
         )
 
+        # NOTE: a workaround for the DocumentJoiner and BranchJoiner components, since _consume_component_inputs()
+        # wraps 'documents' in an extra list, so if there's a 3 level deep list, we need to flatten it to 2 levels only
+        # ToDo: investigate why this is needed and if we can remove it
```
sjrl, line 62: I don't know a solution off the top of my head but this is related to our Variadic type. Specifically `Variadic` is a subclass of `Iterable`. So in all of our joiners we have type `Variadic[SomeType]` which basically translates to `Iterable[SomeType]`. And for the DocJoiner this becomes `Iterable[List[Document]]` hence why we get an extra wrapped list. 

My guess we will need to use the checks `socket.is_variadic`, `socket.is_greedy` and `is_socket_lazy_variadic` to help solve this case generally.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +311,387 @@ def run(  # noqa: PLR0915, PLR0912
...
+
+        # Define the mapping of types to their deserialization functions
+        _type_deserializers = {
+            "Answer": Answer.from_dict,
+            "ChatMessage": ChatMessage.from_dict,
+            "Document": Document.from_dict,
+            "ExtractedAnswer": ExtractedAnswer.from_dict,
+            "GeneratedAnswer": GeneratedAnswer.from_dict,
+            "SparseEmbedding": SparseEmbedding.from_dict,
+        }
```
sjrl, line 533: Similar to what we do for loading components couldn't we first import the `_type` check if it has a `from_dict` attribute and if so then call that? This way we could avoid needing a hard coded list right? [Here](https://github.com/deepset-ai/haystack/blob/0d6a3925068b0df821c295c1a8625911b72c38db/haystack/core/pipeline/base.py#L191-L215) is how we do this for loading components in `base.py`

Would also solve the case for custom components.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +311,387 @@ def run(  # noqa: PLR0915, PLR0912
...
+    @staticmethod
+    def load_state(file_path: Union[str, Path]) -> Dict[str, Any]:
+        """
+        Load a saved pipeline state.
+
+        :param file_path: Path to the state file
+        :returns:
+            Dict containing the loaded state
+        """
+        import json
```
sjrl, line 605: Move import to top of file

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +311,387 @@ def run(  # noqa: PLR0915, PLR0912
...
+
+        file_path = Path(file_path)
+
+        try:
+            with open(file_path, "r", encoding="utf-8") as f:
+                state = json.load(f)
+            Pipeline._validate_resume_state(state=state)
+            logger.info(f"Successfully loaded pipeline state from: {file_path}")
```
sjrl, line 613: Small thing but it looks like the error handling in this try block is only for the lines
```python
            with open(file_path, "r", encoding="utf-8") as f:
                state = json.load(f)
```
so I think it would make sense to take out this line
```python
            Pipeline._validate_resume_state(state=state)
            logger.info(f"Successfully loaded pipeline state from: {file_path}")
```
and move it to after the error handling. Then we could add a new try except block if we wanted to handle the `PipelineRuntimeError` in any special way.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +311,387 @@ def run(  # noqa: PLR0915, PLR0912
...
+        :param resume_state: The saved state to validate.
+        """
+
+        pipeline_state = resume_state["pipeline_state"]
+        valid_components = set(self.graph.nodes.keys())
+
+        # Check if the ordered_component_names are valid components in the pipeline
+        missing_ordered = set(pipeline_state["ordered_component_names"]) - valid_components
+        if missing_ordered:
+            raise PipelineRuntimeError(
```
sjrl, line 637: As a heads up `PipelineRuntimeError` recently got refactored in this [PR](https://github.com/deepset-ai/haystack/pull/9182) which could change how you call it.

haystack_experimental/core/errors.py
```diff
@@ -0,0 +1,59 @@
...
+class SerializationError(Exception):
+    pass
+
+
+class PipelineBreakpointException(Exception):
+    """
+    Exception raised when a pipeline breakpoint is triggered.
+    """
+
+    def __init__(self, message, component=None, state=None, results=None):
```
sjrl, line 55: let's add type hints here as well

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -21,11 +25,17 @@ class Pipeline(PipelineBase):
     Orchestrates component execution according to the execution graph, one after the other.
     """
 
-    def _run_component(
+    ordered_component_names: list[str]
```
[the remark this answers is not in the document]
sjrl: Also for now, let's not mix typing types (e.g. Dict) with builtin types (e.g. dict). I'd say let's consistently use typing types (so Dict) here.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -99,7 +138,12 @@ def _run_component(
             return cast(Dict[Any, Any], component_output)
 
     def run(  # noqa: PLR0915, PLR0912
-        self, data: Dict[str, Any], include_outputs_from: Optional[Set[str]] = None
+        self,
+        data: Dict[str, Any],
+        include_outputs_from: Optional[Set[str]] = None,
+        breakpoints: Optional[Set[Tuple[str, Optional[int]]]] = None,
+        resume_state: Optional[Dict[str, Any]] = None,
+        debug_path: Optional[str] = None,
```
sjrl, line 146: I believe this should be updated to 
```suggestion
        debug_path: Optional[str, Path] = None,
```

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -21,11 +25,17 @@ class Pipeline(PipelineBase):
     Orchestrates component execution according to the execution graph, one after the other.
     """
 
-    def _run_component(
+    ordered_component_names: list[str]
+    original_input_data: dict[str, Any]
+    resume_state: Optional[Dict[str, Any]] = None
+    debug_path: Optional[Union[str, Path]] = None
```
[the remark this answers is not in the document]
davidsbatista: There's not `__init__()`; this class extends Base and only has `run()` and `_run_component()`. We could add it to the `Base` instead, and then they would also be available to the Async version of the Pipeline. 

I did it here at the time so that the changes would be self-contained in a single class.

haystack_experimental/core/pipeline/pipeline.py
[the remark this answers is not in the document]
davidsbatista: I would rather leave them with built-in and also start using the built-in in new code. Currently the ruff config looks like this:

```ini
[tool.ruff]
line-length = 120
target-version = "py38"
exclude = ["test", ".github", "examples"]
```

it should by py39, since we dropped 38 support, and as soon as we change it we will have to change all the typing to use built-in types instead of types from Typing.

haystack_experimental/core/pipeline/pipeline.py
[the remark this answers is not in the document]
sjrl: Okay I see. You could also add an init to this class if you’d like and just also call super init within it. But also fine to move this to base.

haystack_experimental/core/pipeline/pipeline.py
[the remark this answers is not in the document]
sjrl: Right I see what you’re saying. I think @julian-risch and I discussed that we would tackle the type update separately since we have until October for it to be fully deprecated.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -229,45 +303,382 @@ def run(  # noqa: PLR0915, PLR0912
...
+        if isinstance(value, ChatMessage):  # noqa: SIM102
+            if "usage" in value.meta:  # noqa: SIM102
+                value.meta["usage"].pop("completion_tokens_details", None)
+                value.meta["usage"].pop("prompt_tokens_details", None)
+
+        if isinstance(value, GeneratedAnswer):  # noqa: SIM102
+            if value.meta and "usage" in value.meta:  # noqa: SIM102
+                value.meta.pop("usage", None)
+
+        return value
```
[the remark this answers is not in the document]
Amnah199: Opened an issue for this: https://github.com/deepset-ai/haystack/issues/9200

haystack_experimental/core/pipeline/pipeline.py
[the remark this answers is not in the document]
Amnah199: Moving these to base `__init__`.

haystack_experimental/core/pipeline/pipeline.py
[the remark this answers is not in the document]
Amnah199: Updated to stick with typing types (so Dict)  for now.

haystack_experimental/core/pipeline/pipeline.py
davidsbatista: When I bumped into this issue, I noticed that the function that runs before: `_consume_component_inputs()` adds that extra list and it's only for these two joiners Document and Branch (i.e.: the `ListJoiner` and `StringJoiner` are fine)

sjrl: Huh how very odd. Why would ListJoiner and StringJoiner be treated differently here?

davidsbatista: I'm still investigating this and trying out another approach. Anyway, dC could already test this first version of the breakpoints even with this ugly fix.

davidsbatista:
I'm going to merge this into the "temporary" up branch `adding-pipeline-for-easy-PR-review` and I will open one against main from there

PR #271, opened 2025-04-11
feat: adding pipeline breakpoints

davidsbatista:
### Related Issues

- adding pipeline breakpoints
- see https://github.com/deepset-ai/haystack-experimental/pull/269 for more details

### How did you test it?

- unit tests, integration test
- manual tests

### Checklist

- I documented my code


test/core/pipeline/test_pipeline_breakpoints_rag_hybrid.py
```diff
@@ -0,0 +1,153 @@
...
+        """Create and populate a document store for testing."""
+        documents = [
+            Document(content="My name is Jean and I live in Paris."),
+            Document(content="My name is Mark and I live in Berlin."),
+            Document(content="My name is Giorgio and I live in Rome."),
+        ]
+
+        document_store = InMemoryDocumentStore()
+        doc_writer = DocumentWriter(document_store=document_store, policy=DuplicatePolicy.SKIP)
+        doc_embedder = SentenceTransformersDocumentEmbedder(
```
julian-risch, line 38: We should mock the TransformersSimilarityRanker, SentenceTransformersDocumentEmbedder and SentenceTransformersTextEmbedder so that they do not download models and that they do not require the sentence-transformers dependency and also do not rquire transformers with torch.

julian-risch: Similarly, we should mock the OpenAIChatGenerator and OpenAIGenerator so that tests can be much faster

haystack_experimental/core/__init__.py
```diff
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from .pipeline.base import ComponentPriority, PipelineBase
 from .super_component import SuperComponent
 
-_all_ = ["SuperComponent"]
+_all_ = ["ComponentPriority", "PipelineBase", "SuperComponent"]
```
julian-risch, line 8: We don't need to expose ComponentPriority and PipelineBase here or do we? We don't expect users to work with those

haystack_experimental/core/pipeline/errors.py
```diff
@@ -0,0 +1,91 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Optional, Type
+
+
+class PipelineError(Exception):
```
julian-risch, line 8: We already have haystack_experimental/core/errors.py

julian-risch (changes requested):
Looks quite good to me already. I suggest mainly some changes to the tests and errors.py seems duplicated.

haystack_experimental/core/__init__.py
davidsbatista: this was just copy-paste verbatim from main haystack to have a fully functional pipeline - I will adjust

test/core/pipeline/test_pipeline_breakpoints_loops.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            _ = validation_loop_pipeline.run(data, breakpoints={(component, 0)}, debug_path=str(output_directory))
+        except PipelineBreakpointException:
+            pass
+
+        all_files = list(output_directory.glob("*"))
+        file_found = False
+        for full_path in all_files:
+            # windows paths are not POSIX
+            f_name = str(full_path).split("\\")[-1] if os.name == "nt" else str(full_path).split("/")[-1]
```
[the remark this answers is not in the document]
julian-risch: I noticed we are using a similar block of code in multiple places and the main difference seems to be the pipeline:
```python
all_files = list(output_directory.glob("*"))
file_found = False
for full_path in all_files:
    # windows paths are not POSIX
    f_name = str(full_path).split("\\")[-1] if os.name == "nt" else str(full_path).split("/")[-1]
    if str(f_name).startswith(component):
        file_found = True
        resume_state = Pipeline.load_state(full_path)
        result = hybrid_rag_pipeline.run(data, breakpoints=None, resume_state=resume_state)
```
You could consider making it reusable.

julian-risch (changes requested):
We should remove the dependencies from pyproject.toml that are no longer needed in the tests. After that, I approve merging this PR taking into account it's experimental. My other comments don't have to be addressed before merging.

When we later merge this into the haystack repo, we should reuse `mock_openai_chat_generator` instead of creating it multiple times. I understand that right now we return different replies so I am okay with keeping it as is in haystack-experimental. Other parts of the tests could be simplified too, such as the mock_run of the ranker https://github.com/deepset-ai/haystack-experimental/pull/271/files#diff-e44ef907c05488db310dbbea4b3d25ef2460679547b9aa2465115804f8e7432fR94 or the file path parsing with `os.name == "nt"`.

test/core/pipeline/test_pipeline_breakpoints_loops.py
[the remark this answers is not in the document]
davidsbatista: they are not exactly the same, but I can try something

julian-risch (approved):
LGTM! 👍

PR #272, opened 2025-04-11
feat: Add a `super_component` decorator

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9175

### Proposed Changes:

 
 
- Creates a base `_SuperComponent` class that contains everything except `to_dict` and `from_dict`. Then in `SuperComponent` we add the default `to_dict` and `from_dict` methods. This should allow users to still directly use `SuperComponent` and allows us to use the `_SuperComponent` base class in the `super_component` decorator where we don't want `to_dict` or `from_dict` to be set. 
- Adds a `super_component` decorator that still preserves the original class information including the signature of the init method. 
- Updated the existing pre-built super components to use the decorator and removed the super().__init__ and to_dict and from_dict where appropriate. 

### How did you test it?

- Existing tests

### Notes for the reviewer

Putting this in experimental for now so we can see how it would effect pre-built super components.

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/core/super_component/super_component.py
+++ b/haystack_experimental/core/super_component/super_component.py
@@ -7,10 +7,11 @@ import inspect
 from types import new_class
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from haystack import component, logging, Pipeline
+from haystack import Pipeline, component, logging
 from haystack.core.pipeline.async_pipeline import AsyncPipeline
 from haystack.core.pipeline.utils import parse_connect_string
 from haystack.core.serialization import default_from_dict, default_to_dict, generate_qualified_class_name
+
 from haystack_experimental.core.super_component.utils import _delegate_default, _is_compatible
 
 logger = logging.getLogger(__name__)
--- a/haystack_experimental/core/super_component/super_component.py
+++ b/haystack_experimental/core/super_component/super_component.py
@@ -473,31 +473,43 @@ def super_component(cls):
     - input_mapping: Dictionary mapping component inputs to pipeline inputs (optional)
     - output_mapping: Dictionary mapping pipeline outputs to component outputs (optional)
     """
+    # Create a new class that inherits from SuperComponent and the original class
+    new_cls_name = cls.__name__
+
     # Keep a reference to the original __init__
     original_init = cls.__init__
 
-    # Define a new __init__ that wraps the original
-    def __init_wrapper__(self, *args, **kwargs):
-        # Call the original __init__ to set up pipeline and mappings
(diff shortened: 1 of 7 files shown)


sjrl:
Hey @julian-risch could I ask you for a review of this? At least of general approach and if you can see anything immediately wrong like the new inheritance structure of SuperComponent?

haystack_experimental/super_components/converters/multi_file_converter.py
```diff
@@ -20,7 +19,7 @@
 from haystack.components.joiners import DocumentJoiner
 from haystack.components.routers import FileTypeRouter
 
-from haystack_experimental.core.super_component import SuperComponent
+from haystack_experimental.core.super_component import super_component
```
julian-risch, line 22: Just a comment: Similar to how users can call from `haystack import component`, I'd say we should also enable them to call `haystack import super_component ` if possible. No need to enable `from haystack_experimental import super_component` here but that's how I imagine it.

test/super_components/converters/test_multi_file_converter.py
```diff
@@ -22,7 +22,6 @@ def test_init_default_params(self, converter):
         """Test initialization with default parameters"""
         assert converter.encoding == "utf-8"
         assert converter.json_content_key == "content"
-        assert isinstance(converter, SuperComponent)
```
julian-risch, line 25: `assert isinstance(converter, Component)` should work to at least check that `converter` implements the `Component` protocol.

haystack_experimental/core/super_component/super_component.py
```diff
@@ -2,40 +2,108 @@
...
+class _SuperComponent:
     """
     A class for creating super components that wrap around a Pipeline.
 
     This component allows for remapping of input and output socket names between the wrapped pipeline and the
     SuperComponent's input and output names. This is useful for creating higher-level components that abstract
     away the details of the wrapped pipeline.
-    """
 
+    ### Usage example
```
julian-risch, line 40: It's great that we don't cause any breaking change in this usage example. We just need to move it to `SuperComponent` instead of `_SuperComponent`.

julian-risch (commented):
Usage of @super_component with MultiFileConverter is exactly as I imagined and it’s great to see how we don’t need to define to_dict/from_dict any longer.
I couldn't spot any issues in the code. You could do the changes in a PR in Haystack core directly. And then we can move the ready-made components in a separate PR.

sjrl:
Closing this since will directly make a PR into Haystack

PR #276, opened 2025-04-14
Revert internal call to pipeline_load state

Amnah199:
Instead of passing the state file path, we allow the user to call `load_state` and pass the state to `run` method to resume the pipeline. This was requested by Thomas.


davidsbatista (approved):
LGTM!

PR #286, opened 2025-04-23
feat: `ImageContent` dataclass

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack/issues/9258

### Proposed Changes:

- introduce `ImageContent` dataclass to represent images in `ChatMessage`
- refactor `ChatMessage.from_user` method to support an ordered list of texts and images

### How did you test it?
CI, several new/updated tests

### Notes for the reviewer
- I tried to introduce as less code as possible and reuse `ChatMessage` from Haystack

TODO in future PRs:
- add convenience methods (`from_file`, `from_url`)
- adapt OpenAI conversion methods

### Checklist

- I documented my code


haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@dataclass
+class ImageContent:
+    """
+    The image content of a chat message.
+
+    :param base64_image: A base64 string representing the image.
+    :param mime_type: The mime type of the image.
+    :param detail: Optional detail level of the image (only supported by OpenAI). One of "auto", "high", or "low".
```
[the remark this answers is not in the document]
anakin87: I understand your point.

My initial implementation included `provider_options: Dict[str, Any] = field(default_factory=dict)` instead of `detail`.

Then I discovered out that:
- LangChain and LlamaIndex use `detail` (if I remember correctly)
- similar parameters are not supported by major model providers (but they may support them or others in the future)

`provider_options` would be more future-proof but a bit less user-friendly.

Should we choose this solution?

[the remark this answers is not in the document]
sjrl: Okay I see what you're saying. I was under the impression Anthropic also has this, but it looks like from their [docs](https://docs.anthropic.com/en/docs/build-with-claude/vision#example-four-images-across-two-conversation-turns) they just automatically handle the resizing for you if you aren't compliant with one of their listed [sizes](https://docs.anthropic.com/en/docs/build-with-claude/vision#evaluate-image-size). 

I think then it's probably fine to leave it as is with `detail` then.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,187 @@
...
+    mime_type: Optional[str] = None
+    detail: Optional[Literal["auto", "high", "low"]] = None
+    meta: Dict[str, Any] = field(default_factory=dict)
+
+    def __post_init__(self):
+        # mime_type is an important information, so we try to guess it if not provided
+        if not self.mime_type:
+            try:
+                # Attempt to decode the string as base64
+                decoded_image = base64.b64decode(self.base64_image)
```
sjrl, line 39: I believe this operation can get quite slow for large images (more than 10s of MBs). I think it's fine to have, but probably good to put a warning in the docstring saying if mime_type is left blank it will try and guess it by decoding the image.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,187 @@
...
+
+                guess = filetype.guess(decoded_image)
+                if guess:
+                    self.mime_type = guess.mime
+            except:
+                pass
```
sjrl, line 45: I think I might log a warning here if there was a decoding issue. Could indicate something wrong with the base64 encoded string right?

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,187 @@
...
+    def __post_init__(self):
+        # mime_type is an important information, so we try to guess it if not provided
+        if not self.mime_type:
+            try:
+                # Attempt to decode the string as base64
+                decoded_image = base64.b64decode(self.base64_image)
+
+                guess = filetype.guess(decoded_image)
+                if guess:
+                    self.mime_type = guess.mime
```
sjrl, line 43: Do you think it'd make sense to add a logger message here the in situation no mimetype was found?

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -0,0 +1,187 @@
...
+        content: List[Dict[str, Any]] = []
+        for part in self._content:
+            if isinstance(part, TextContent):
+                content.append({"text": part.text})
+            elif isinstance(part, ToolCall):
+                content.append({"tool_call": asdict(part)})
+            elif isinstance(part, ToolCallResult):
+                content.append({"tool_call_result": asdict(part)})
+            elif isinstance(part, ImageContent):
+                content.append({"image": asdict(part)})
```
sjrl, line 182: I wonder if it would also make sense to have a way to serialize `ImageContent` by pointing to a `file_path` instead of it needing to be a base64_string. 

Similar to how you added a `from_file_path` utility method to `ImageContent` in your PoC I wonder if we should allow users to serialize and deserialize with this pointer instead. 

I'm thinking this b/c storing ChatMessages could become quite large in in disk if we can only store the base64 string and not optionally set a pointer instead. What do you think?

sjrl:
Looks good @anakin87! I have a few minor comments and only one larger [one](https://github.com/deepset-ai/haystack-experimental/pull/286#discussion_r2058377119).

haystack_experimental/dataclasses/chat_message.py
anakin87: Whenever we templatize `ImageContent` (e.g. in `ChatPromptBuilder`), this is expected, so using `logger.warning` might become noisy. Should we use `logger.info`?

haystack_experimental/dataclasses/chat_message.py
anakin87: makes sense. I'll add this information

haystack_experimental/dataclasses/chat_message.py
anakin87: maybe here it makes more sense, since it will be emitted only if the base64 image is valid and we cannot guess this relevant information. I'll add a warning.

haystack_experimental/dataclasses/chat_message.py
anakin87: Currently, I have the impression that serializing (and storing `ChatMessage`) is mostly used for `ChatPromptBuilder` and for multimodal inputs I would like to avoid so (as illustrated in the POC).

Did I miss something? Can you think of other use cases?

haystack_experimental/dataclasses/chat_message.py
anakin87: https://github.com/deepset-ai/haystack-experimental/pull/286/commits/e1eafc3a7fd98727f6f0109b4dd6bce263813d9f

haystack_experimental/dataclasses/chat_message.py
anakin87: https://github.com/deepset-ai/haystack-experimental/pull/286/commits/e1eafc3a7fd98727f6f0109b4dd6bce263813d9f

haystack_experimental/dataclasses/chat_message.py
sjrl: > Can you think of other use cases?

I think some up and coming ones are short-term and long-term memory for Agent. Where we will probably revisit the ChatMessageWriter and ChatMessageRetriever in haystack-experimental. 

Additionally dC would have to decide how to handle this since we are basically writing the contents of ChatMessage to a dB to retain chat history. 

I think maybe a good alternative/workaround could be to handle in the `ChatMessageWriter` and `Retriever` how we'd like to handle ImageContent within ChatMessages. I.e. we could control there whether to skip writing the base64 string to the DB, but then I'd think we should store the file_path to the original image in the meta-field of ImageContent. What do you think?

haystack_experimental/dataclasses/chat_message.py
sjrl: Ahh I see, you're saying using ImagConent with a Jinja2 template in ChatPromptBuilder will cause this to throw an exception. I agree that it'd become too noisy so probably best to leave alone for now. 

Maybe we can come back to this later if we think of a way of avoiding this when templatizing (like with your new chat template proposal).

anakin87: OK, let's reconsider this aspect in the future.

About the new chat template: even if it has its advantages, it still requires internal templatization to some extent, so it would not improve this aspect (differently from what I initially thought).

haystack_experimental/dataclasses/chat_message.py
anakin87: Thanks for the clarification, makes sense.

Handling this at the `ChatMessageWriter` and `Retriever` level sounds like a practical approach, and it helps keep the scope of the current multimodal work focused for now.

Storing a `file_path` in the `meta` field of `ImageContent` is feasible and would help keep the `ChatMessage` serialization logic simpler at this stage.

As we move forward, we can revisit/adjust this strategy.

WDYT?

sjrl: Sounds good to me!

sjrl (approved):
Looks good!

PR #290, opened 2025-04-25
feat: Add `ImageFileToImageContent` converter

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9262

### Checklist

- I documented my code


haystack_experimental/components/converters/pdf_to_image.py
```diff
@@ -0,0 +1,139 @@
...
+        detail: Optional[Literal["auto", "high", "low"]] = None,
+        downsize: bool = False,
+        page_range: Optional[List[Union[str, int]]] = None,
+    ):
+        """
+        Create the PDFToImageContent component.
+
+        :param detail: Optional detail level of the image (only supported by OpenAI). One of "auto", "high", or "low".
+        :param downsize: If True, the image will be downscaled to the specified detail level.
+        :param page_range: A range of pages to extract metadata from. For example, page_range=['1', '3'] will extract
```
anakin87, line 38: If I understand well, `page_range` indicates the pages to convert to images.
E.g., if I have a PDF of 100 pages and I specify `page_range=['1', '3']`, I will get a list of 2 `ImageContent` objects.

If I am correct, we should fix the docstrings accordingly and communicate this aspect.

haystack_experimental/components/converters/image_utils.py
```diff
@@ -0,0 +1,278 @@
...
+
+    # Don't resize image if it's already within requirements
+    if image.width <= new_width and image.height <= new_height:
+        return image
+
+    resized_image = image.resize((new_width, new_height))
+    return resized_image
+
+
+def read_image_from_pdf(
```
anakin87, line 185:
```suggestion
def read_images_from_pdf(
```

this would clarify the goal of this function

anakin87 (commented):
I agree with the overall design.

I have left some initial comments.

It might also be worth trying to simplify/choose clearer names for the functions in image_utils.

(Splitting this PR into 2 PRs could have helped the review process.)

docs/pydoc/config/converters_api.yml
```diff
@@ -0,0 +1,31 @@
...
+    skip_empty_modules: true
+  - type: smart
+  - type: crossref
+renderer:
+  type: haystack_pydoc_tools.renderers.ReadmeCoreRenderer
+  excerpt: Various converters to transform data from one format to another.
+  category_slug: experiments-api
+  title: Converters
+  slug: experimental-converters-api
+  order: 20
```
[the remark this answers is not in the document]
sjrl: Ahh thanks I mostly copied this directly from Haystack.

.github/workflows/tests.yml
```diff
@@ -70,6 +70,23 @@ jobs:
...
+        if: matrix.os == 'macos-latest'
+        run: brew install poppler
+
+      - name: Install Poppler-utils (Windows)
+        if: matrix.os == 'windows-latest'
+        run: |
+          Invoke-WebRequest -Uri https://github.com/oschwartz10612/poppler-windows/releases/download/v24.08.0-0/Release-24.08.0-0.zip -OutFile poppler.zip
+          Expand-Archive poppler.zip -DestinationPath poppler
+          echo "${{ github.workspace }}\poppler\poppler-24.08.0\Library\bin" | Out-File -Append -FilePath $env:GITHUB_PATH
+          echo "Poppler installed."
```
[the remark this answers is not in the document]
sjrl: Ahh thanks for the heads up. 

I'll definitely be happy to look for alternatives as well.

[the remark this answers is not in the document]
sjrl: A potential alternative could be to use https://github.com/pypdfium2-team/pypdfium2 which uses [PDFium](https://pdfium.googlesource.com/pdfium/+/HEAD/docs/getting-started.md) from Google under-the-hood. This doesn't help with the installation process since we will still need to manually install pre-built binaries from [here](https://github.com/bblanchon/pdfium-binaries) but should be permissive with licensing. 

PDFium's license looks to be Apache 2.0 https://pdfium.googlesource.com/pdfium/+/refs/heads/main/LICENSE
pypdfium2 is also under permissive licensing https://github.com/pypdfium2-team/pypdfium2?tab=readme-ov-file#licensing

sjrl:
@anakin87 I could still split this up into two PRs one for the Image conversion and support (e.g. for png, jpg, etc.) and then a separate one for the PDF converters since there is a bigger topic there about licensing. WDYT?

haystack_experimental/components/converters/file_to_image.py
```diff
@@ -0,0 +1,106 @@
...
+_EMPTY_BYTE_STRING = b""
+
+
+@component
+class ImageFileToImageContent:
+    """
+    Converts image files to ImageContent objects.
+
+    :param detail: Optional detail level of the image (only supported by OpenAI). One of "auto", "high", or "low".
+    :param downsize: If True, the image will be downscaled to the specified detail level.
```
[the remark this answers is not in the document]
sjrl: I've improved the docstring to reflect this. Let me know if the new one needs anymore tweaking.

haystack_experimental/components/converters/image_utils.py
sjrl: I've updated the name to be `convert_pdf_to_images`

haystack_experimental/components/converters/pdf_to_image.py
sjrl: I've updated the docstring. Let me know if it's more clear!

sjrl:
@anakin87 this is ready for another review in regards to everything but the poppler dependency. I'm still investigating the alternative I mentioned [here](https://github.com/deepset-ai/haystack-experimental/pull/290#discussion_r2063475838)

haystack_experimental/components/converters/file_to_image.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    Converts image files to ImageContent objects.
+    """
+
+    def __init__(self, *, detail: Optional[Literal["auto", "high", "low"]] = None, downsize: bool = False):
+        """
+        Create the ImageFileToImageContent component.
+
+        :param detail: Optional detail level of the image (only supported by OpenAI). One of "auto", "high", or "low".
+        :param downsize: If True, resizes the image to fit within the specified dimensions while maintaining aspect
```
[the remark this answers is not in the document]
sjrl: That's a fair point. I'd be happy to allow `downsize` to be a runtime param as well. 

Also on this point. Do you think it would make sense to have a separate size param separate from `detail` that controls the resizing?

haystack_experimental/components/converters/file_to_image.py
```diff
@@ -0,0 +1,111 @@
...
+            ratio. This reduces file size, memory usage, and processing time, which is beneficial when working with
+            models that have resolution constraints or when transmitting images to remote services.
+        """
+        self.detail = detail
+        self.downsize = downsize
```
sjrl, line 38: @anakin87 do you think it would be worth to add a `pillow_import.check()` here if `detail` and `downsize` are provided? That's the condition for needing Pillow installed. 

The only downside I see is if the user overwrites these values at runtime such that Pillow is not needed anymore (e.g. downsize=False).

anakin87: ok, what about adding a similar check in the `run` method?

haystack_experimental/components/converters/file_to_image.py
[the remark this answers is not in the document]
anakin87: I am trying to reflect on this point, but would be interested to hear your opinion.

If we leave things as they are now, we should clarify that `detail` role is twofold:
- in combination with `downsize=True`, resizes the image (already well explained)
- also corresponds to the `detail` parameter of OpenAI API, that impacts the number of tokens used.

haystack_experimental/components/converters/file_to_image.py
sjrl: added

.github/workflows/tests.yml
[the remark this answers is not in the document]
sjrl: @anakin87 I think I will leave poppler in for now, but open this as another issue to explore/consider

haystack_experimental/components/converters/file_to_image.py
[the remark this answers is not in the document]
anakin87: I trust you on this aspect, but adding a `size` parameter would make the interface much cleaner and clearer, in my opinion.

Perhaps the `size` parameter could completely replace `downscale`. What do you think?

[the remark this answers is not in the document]
sjrl: Yeah that's a good idea

haystack_experimental/components/converters/image_utils.py
```diff
@@ -0,0 +1,301 @@
...
+        maintaining aspect ratio. This reduces file size, memory usage, and processing time, which is beneficial
+        when working with models that have resolution constraints or when transmitting images to remote services.
+
+    :returns:
+        A tuple (mime_type, base64_str), where:
+        - mime_type (str): The MIME type of the encoded image, determined from the original data or image content.
+          Defaults to 'image/jpeg' if the type cannot be reliably identified.
+        - base64_str (str): The base64-encoded string representation of the (optionally resized) image.
+    """
+    if size is None:
```
anakin87, line 83: there might be an inconsistency related to`mime_type` here:
- if `size` is None, we return `bytestream.mime_type` which can be `None` if the original source was a `ByteStream` (not a path)
- otherwise, we try to infer it and return `"image/jpeg"` if we fail to infer it (this is also done because we need to specify a `form` in `image.save`)

(I'm not sure of my reasoning because I see that mypy is not raising any error.)

For consistency, I would prefer to return `"image/jpeg"` also when `bytestream.mime_type` is `None`. Makes sense?

haystack_experimental/components/converters/image_utils.py
```diff
@@ -0,0 +1,301 @@
...
+        return image
+
+    resized_image = image.resize((new_width, new_height))
+    return resized_image
+
+
+def convert_pdf_to_images(
+    bytestream: ByteStream,
+    page_range: Optional[List[int]] = None,
+    size: Optional[Union[Tuple[int, int]]] = None,
```
anakin87, line 205: is `Union` needed here?

anakin87 (commented):
I left some minor comments.

I've also spent some time trying to figure out how to simplify the handling of `mime_type' in utility functions, but haven't found a better solution.

One thing to keep in mind: I see that currently lazy import checks seem a bit scattered and sometimes we are not using forward references in types. Perhaps you can check this aspect.

haystack_experimental/components/converters/image_utils.py
sjrl: no, will remove

haystack_experimental/components/converters/image_utils.py
anakin87: Ok, I understand and seem to make sense.

What I would do:
- change the return type of `encode_image_to_base64` to `Tuple[Optional[str], str]`
- try to explain this aspect in the docstring
- (maybe) `if bytestream.mime_type is None`, show a warning similar to the `if not inferred_mime_type` below

sjrl: Done! I've also added a warning message. I think it could be improved but I'm not entirely sure how. Mostly surrounding how to tell the user how to provide the missing mime type.

anakin87: thanks!

haystack_experimental/components/converters/pdf_to_image.py
```diff
@@ -0,0 +1,140 @@
...
+            If None, all pages in the PDF will be converted. Pages outside the valid range (1 to number of pages)
+            will be skipped with a warning. For example, page_range=[1, 3] will convert only the first and third
+            pages of the document. It also accepts printable range strings, e.g.:  ['1-3', '5', '8', '10-12']
+            will convert pages 1, 2, 3, 5, 8, 10, 11, 12.
+        """
+        self.detail = detail
+        self.size = size
+        self.page_range = page_range
+        pypdf_and_pdf2image_import.check()
+
```
anakin87, line 53: do we also need `pillow_import.check()` here?

pyproject.toml
```diff
@@ -69,7 +69,8 @@ extra-dependencies = [
   "python-pptx", # MultiFileConverter
   "trafilatura", # MultiFileConverter
   "transformers[torch,sentencepiece]>=4.51.1,<4.52", # same as in haystack-ai
-
+  "pdf2image", # for pdf to image conversion in FileToImageContent
+  "pillow", # for image processing in FileToImageContent
```
anakin87, line 73:
```suggestion
  "pdf2image", # for pdf to image conversion in PDFToImageContent
  "pillow", # for image processing in ImageFileToImageContent and PDFToImageContent
```

anakin87 (approved):
Looks good!

- I left two minor comments
- We talked about putting these components in `image_converters` instead of `converters`. Do you want to do that now?

haystack_experimental/components/converters/pdf_to_image.py
sjrl: No because pdf2image will download PIL when installing pdf2image.

PR #292, opened 2025-04-30
feat: multimodal support in `OpenAIChatGenerator`

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9259

### Proposed Changes:
- Update the experimental `ChatMessage.to_openai_dict_format` to support user messages with images
- Introduce an experimental `OpenAIChatGenerator`, which uses the experimental `ChatMessage` (no other changes in the component)
- Partly unrelated: make `ChatMessage.from_user` raise an error if none of the provided content parts is textual

### How did you test it?
- improved/added tests for `ChatMessage` (here we test the core logic of message conversion)
- to ensure that everything works on the experimental `OpenAIChatGenerator`, I copied a substantial number of tests from Haystack and added a new test for multimodality

### Notes for the reviewer
- Don't be scared by the size of this PR. It consists mostly of tests.
- Changes in `ChatMessage.to_openai_dict_format` (compared to the Haystack version)  can be inspected in this commit: https://github.com/deepset-ai/haystack-experimental/pull/292/commits/916a66d6a84647212ba989a867559c47598a51ea

### Checklist

- I documented my code


haystack_experimental/dataclasses/chat_message.py
```diff
@@ -163,6 +164,8 @@ def from_user(
             content = [TextContent(text=text)]
         elif content_parts is not None:
             content = [TextContent(el) if isinstance(el, str) else el for el in content_parts]
+            if not any(isinstance(el, TextContent) for el in content):
+                raise ValueError("The user message must contain at least one textual part.")
```
sjrl, line 168: Out of curiosity why is this necessary? Is it true that all LLM providers require some text and can't just be an image (or e.g. audio?)

docs/pydoc/config/generators_api.yml
```diff
@@ -0,0 +1,30 @@
...
+  title: Generators
+  slug: generators-api
+  order: 20
+  markdown:
+    descriptive_class_title: false
+    classdef_code_block: false
+    descriptive_module_title: true
+    add_method_class_prefix: true
+    add_member_class_prefix: false
+    filename: generators_api.md
```
sjrl, line 30: Just for my own knowledge. I've seen other pydoc configs add `experimental_` to the beginning of filename. Is that not necessary?

docs/pydoc/config/generators_api.yml
```diff
@@ -0,0 +1,30 @@
...
+    skip_empty_modules: true
+  - type: smart
+  - type: crossref
+renderer:
+  type: haystack_pydoc_tools.renderers.ReadmeCoreRenderer
+  excerpt: Enables text generation using LLMs.
+  category_slug: haystack-api
+  title: Generators
+  slug: generators-api
+  order: 20
```
sjrl, line 23: Also I believe this has the same order number as dataclasses

docs/pydoc/config/generators_api.yml
anakin87: You are right. Fixed in https://github.com/deepset-ai/haystack-experimental/pull/292/commits/d45a03b111c0f001869b14b1706a4faf70759775.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -193,3 +196,77 @@ def to_dict(self) -> Dict[str, Any]:
...
+        text_contents = self.texts
+        tool_calls = self.tool_calls
+        tool_call_results = self.tool_call_results
+
+        if not text_contents and not tool_calls and not tool_call_results:
+            raise ValueError(
+                "A `ChatMessage` must contain at least one `TextContent`, `ToolCall`, or `ToolCallResult`."
+            )
+        if len(text_contents) + len(tool_call_results) > 1:
+            raise ValueError("A `ChatMessage` can only contain one `TextContent` or one `ToolCallResult`.")
```
sjrl, line 213: General question. While working on the Bedrock Integration I noticed apparently it's possible to have mixed contents from one response. So basically a text and tool call result, of which there is a bug on how to handle this opened [here](https://github.com/deepset-ai/haystack-core-integrations/issues/1681). 

So the question I wanted to ask is this ValueError generally true or only true for OpenAI? If it's generally true it may be good to add this check in the `from_assistant` method.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -193,3 +196,77 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+        # Add name field if present
+        if self._name is not None:
+            openai_msg["name"] = self._name
+
+        # user message
+        if openai_msg["role"] == "user":
+            if len(self._content) == 1:
+                openai_msg["content"] = self.text
+                return openai_msg
```
sjrl, line 225: Just clarifying this works and doesn't need an isinstance check on what `self._content` is because if it's length 1 it has to be text?

haystack_experimental/dataclasses/chat_message.py
anakin87: I think yes, if we are talking about models that accept both text and images. I checked OpenAI, Anthropic, Google, and Ollama.

docs/pydoc/config/generators_api.yml
sjrl: ahh good catch, was on a old version locally

haystack_experimental/dataclasses/chat_message.py
sjrl: Okay sounds fine to leave then.

sjrl: I wonder if this becomes true when we eventually get to audio. But we can handle that then!

haystack_experimental/dataclasses/chat_message.py
sjrl: I'd say it'd be good to update the error message to reflect that this is an OpenAI restraint. I recognize users are already calling `to_openai_...`, but I think it would still be helpful.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -193,3 +196,77 @@ def to_dict(self) -> Dict[str, Any]:
...
+            for part in self._content:
+                if isinstance(part, TextContent):
+                    content.append({"type": "text", "text": part.text})
+                elif isinstance(part, ImageContent):
+                    image_item: Dict[str, Any] = {
+                        "type": "image_url",
+                        # If no MIME type is provided, default to JPEG.
+                        # OpenAI API appears to tolerate MIME type mismatches.
+                        "image_url": {"url": f"data:{part.mime_type or 'image/jpeg'};base64,{part.base64_image}"},
+                    }
```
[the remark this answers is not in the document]
anakin87: I initially decided to exclude this use case for a few reasons:
- keeping `ImageContent` simpler
- passing a URL is supported in OpenAI and Anthropic, but some other providers do not provide this feature ([Google](https://ai.google.dev/gemini-api/docs/image-understanding?lang=python#inline-image), [Ollama](https://github.com/ollama/ollama-python/blob/main/examples/multimodal-chat.py))

I am open to discussion.

haystack_experimental/dataclasses/chat_message.py
anakin87: Good idea. Done in https://github.com/deepset-ai/haystack-experimental/pull/292/commits/26a3b4a30b3c2c9045a5c6ec1ec3cfd6f4f4093d

haystack_experimental/components/generators/chat/openai.py
```diff
@@ -0,0 +1,20 @@
...
+import haystack.components.generators.chat.openai
+from haystack_experimental.dataclasses.chat_message import ChatMessage
+from haystack import component
+
+# Monkey patch the Haystack ChatMessage class with the experimental one. By doing so, we can use the new 
+# `to_openai_dict_format` method, allowing multimodal chat messages.
+haystack.components.generators.chat.openai.ChatMessage = ChatMessage
+
+@component
+class OpenAIChatGenerator(haystack.components.generators.chat.openai.OpenAIChatGenerator):
```
davidsbatista, line 14: Just curious to understand: this is here because there's currently no OpenAIChatGenerator on experimental? do we really need it, couldn't we import it from haystack core?

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -11,6 +11,7 @@
 from haystack import logging
 from haystack.dataclasses import ChatMessage as HaystackChatMessage
 from haystack.dataclasses import ChatRole, TextContent, ToolCall, ToolCallResult
+import json
```
davidsbatista, line 14: nit: I suspect this import should be done before the haystack imports

haystack_experimental/components/generators/chat/openai.py
anakin87: It's a bit tricky, but doing `from haystack_experimental.components.generators.chat import OpenAIChatGenerator` ensures that the line before (where we monkey-patch `ChatMessage` with its experimental counterpart) is always executed.

davidsbatista: ok, got it - I see the point now 👍🏽

haystack_experimental/dataclasses/chat_message.py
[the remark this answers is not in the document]
sjrl: Okay sounds good, let's leave it out for now

sjrl (approved):
Looks good!

PR #293, opened 2025-05-04
fix: preserve initialization parameters in debug state when run params are not supplied 

### LINKED ISSUE #284: bug: `init` Parameters Not Stored in Pipeline State Files
While testing advanced use cases involving breakpoints—such as pausing at a `PromptBuilder` and manually modifying prompt templates in saved states to compare results—I noticed that `init` parameters are not stored in the state files. Instead, they're stored only in the pipeline configuration when the components are first initialized.

This generally works fine, since users typically resume states using the same pipeline with the original component configurations. However, in more advanced scenarios like the one above, this can lead to confusion or even errors, especially when a prompt template is unexpectedly stored as null.

**Suggestion**
We should investigate how to properly persist component init parameters in the pipeline state to prevent unintended overrides or missing values when resuming from saved states.

**Example**
Currently, for a `ChatPromptBuilder`  initialized like this:

```
template = [
        ChatMessage.from_system(
            "You are a helpful AI assistant. Answer the following question based on the given context information only. If the context is empty or just a '\n' answer with None, example: 'None'."
        ),
        ChatMessage.from_user(
            """
            Context:
            {% for document in documents %}
                {{ document.content }}
            {% endfor %}
    
            Question: {{question}}
            """
        )
    ]
rag_pipeline.add_component(instance=ChatPromptBuilder(template=template, required_variables=["question", "documents"]), name="prompt_builder", )    
```

When a breakpoint is set at the `prompt_builder` component, the saved state looks like this:

```
"prompt_builder": {
        "question": "Where does Mark live?",
        "documents": [...],
        "template": null,
        "template_variables": null
      },
```

Amnah199:
### Related Issues

- fixes #284 

### Proposed Changes:
Added functionality to preserve initialization parameters in debug state. When parameters are defined in component initialization but null in the run method, we now automatically restore the original init values in the saved pipeline state for improved debugging.

### DIFF
--- a/haystack_experimental/core/pipeline/pipeline.py
+++ b/haystack_experimental/core/pipeline/pipeline.py
@@ -122,6 +122,7 @@ class Pipeline(PipelineBase):
             span.set_content_tag("haystack.component.input", deepcopy(component_inputs))
             logger.info("Running component {component_name}", component_name=component_name)
             try:
+                print(f"Running component {component_name} with inputs: {component_inputs}")
                 component_output = instance.run(**component_inputs)
             except Exception as error:
                 raise PipelineRuntimeError.from_exception(component_name, instance.__class__, error) from error
--- a/haystack_experimental/core/pipeline/pipeline.py
+++ b/haystack_experimental/core/pipeline/pipeline.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import json
-from copy import deepcopy
+from copy import copy, deepcopy
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Mapping, Optional, Set, Tuple, Union, cast
@@ -89,9 +89,10 @@ class Pipeline(PipelineBase):
             breakpoint_inputs = deepcopy(inputs)
             # we deepcopy the component_inputs to avoid modifying the original inputs
             breakpoint_inputs[component_name] = deepcopy(Pipeline._remove_unserializable_data(component_inputs))
-            if hasattr(component["instance"], "to_dict"):
-                params = component["instance"].to_dict()
-                breakpoint_inputs[component_name]["init_parameters"] = params["init_parameters"]
+            params = copy(component["instance"].__dict__)
+            if any(key == "_template_string" for key in params):
+                params["template"] = params["_template_string"]
+            breakpoint_inputs[component_name]["init_parameters"] = params
             self._check_breakpoints(breakpoints, component_name, component_visits, breakpoint_inputs)
 
         with tracing.tracer.trace(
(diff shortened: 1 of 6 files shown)


haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -83,10 +84,15 @@ def _run_component(  # pylint: disable=too-many-positional-arguments
...
-        breakpoint_inputs = deepcopy(inputs)
-        breakpoint_inputs[component_name] = Pipeline._remove_unserializable_data(component_inputs)
         if breakpoints and not self.resume_state:
+            # add component_inputs to inputs
+            breakpoint_inputs = deepcopy(inputs)
+            # we deepcopy the component_inputs to avoid modifying the original inputs
+            breakpoint_inputs[component_name] = deepcopy(Pipeline._remove_unserializable_data(component_inputs))
+            params = copy(component["instance"].__dict__)
+            if any(key == "_template_string" for key in params):
+                params["template"] = params["_template_string"]
```
Amnah199, line 94: This is another edge case. Using `to_dict()` here strips away class types like `ChatMessage`, which makes deserialization of the state problematic. On the other hand, using `__dict__` relies on class variables being stored with the same names as their constructor parameters.
For example, the template parameter is stored as `self._template_string` in this class, so it needs to be handled. 

This is not the best solution and can fail as we don't have a strict protocol for naming class vars.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -566,6 +571,15 @@ def save_state(
 
         dt = datetime.now()
         file_name = Path(f"{component_name}_{dt.strftime('%Y_%m_%d_%H_%M_%S')}.json")
+        # we store init_parameters for the component with breakpoint in the saved state
```
davidsbatista, line 574: I was a bit confused with some of the comments

```suggestion
        # we store a component init_parameters together with breakpoint in the saved state
```

haystack_experimental/core/pipeline/pipeline.py
davidsbatista: OK, I understand this issue now, and the use case is relevant, i.e.: re-running/resuming partly the pipeline and changing only the prompt.

I would suggest to add a comment summarising what you explained above 

```python
if any(key == "_template_string" for key in params):
    params["template"] = params["_template_string"]
```

so that in the future one can remember why this code is here

davidsbatista (approved):
LGTM! I think we should just comment the edge case you are handling for the `template` in the ChatPromptBuilder - feel free to merge it after.

haystack_experimental/core/pipeline/pipeline.py
Amnah199: My concern is currently we dont have a better workaround and if we want to merge this in core repo, these fixes for edge cases can be very problematic.

Amnah199: Side note, I proposed adding our ser/deseriaization functions as util functions to `haystack` as they are useful in other places. 

Here is the source: https://github.com/deepset-ai/haystack/blob/0190e0497ba7013b27542e1aa6cf150d016fb8d0/haystack/utils/base_serialization.py#L58

PR #294, opened 2025-05-06
feat: add convenience class methods to `Imagecontent`

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack/issues/9258

### Proposed Changes:
- add `ImageContent.from_file_path` class method
- add `ImageContent.from_url` class method

### How did you test it?
CI; several new unit tests

### Notes for the reviewer
- I decided to reuse the existing components in these class methods (`ImageFileToImageContent` and `LinkContentFetcher`) to avoid duplication and reduce maintenance efforts
- One aspect I want to discuss is MIME type: in the current implementation of these class methods, the user is free to set it and in this case we don't infer it from file path/URL content type. To keep the code simple, an alternative would be to remove this parameter entirely and always infer the MIME type. Given a file path or URL, MIME type inference is generally reliable. What do you think?

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/image_converters/file_to_image.py
+++ b/haystack_experimental/components/image_converters/file_to_image.py
@@ -126,7 +126,6 @@ class ImageFileToImageContent:
                 )
                 continue
 
-
             merged_metadata = {**bytestream.meta, **metadata}
             image_content = ImageContent(
                 base64_image=base64_image, mime_type=inferred_mime_type, meta=merged_metadata, detail=resolved_detail
--- a/haystack_experimental/components/image_converters/file_to_image.py
+++ b/haystack_experimental/components/image_converters/file_to_image.py
@@ -126,8 +126,6 @@ class ImageFileToImageContent:
                 )
                 continue
 
-            print(f"metadata: {metadata}")
-            print(f"bytestream.meta: {bytestream.meta}")
 
             merged_metadata = {**bytestream.meta, **metadata}
             image_content = ImageContent(
--- a/haystack_experimental/components/image_converters/file_to_image.py
+++ b/haystack_experimental/components/image_converters/file_to_image.py
@@ -69,11 +69,11 @@ class ImageFileToImageContent:
         :param sources:
             List of file paths or ByteStream objects to convert.
         :param meta:
-            Optional metadata to attach to the documents.
+            Optional metadata to attach to the ImageContent objects.
             This value can be a list of dictionaries or a single dictionary.
-            If it's a single dictionary, its content is added to the metadata of all produced documents.
+            If it's a single dictionary, its content is added to the metadata of all produced ImageContent objects.
             If it's a list, its length must match the number of sources as they're zipped together.
-            For ByteStream objects, their `meta` is added to the output documents.
+            For ByteStream objects, their `meta` is added to the output ImageContent objects.
         :param detail:
             Optional detail level of the image (only supported by OpenAI). One of "auto", "high", or "low".
             This will be passed to the created ImageContent objects.
--- a/haystack_experimental/components/image_converters/file_to_image.py
+++ b/haystack_experimental/components/image_converters/file_to_image.py
@@ -126,6 +126,9 @@ class ImageFileToImageContent:
                 )
                 continue
 
+            print(f"metadata: {metadata}")
+            print(f"bytestream.meta: {bytestream.meta}")
+
             merged_metadata = {**bytestream.meta, **metadata}
             image_content = ImageContent(
                 base64_image=base64_image, mime_type=inferred_mime_type, meta=merged_metadata, detail=resolved_detail
--- a/haystack_experimental/dataclasses/chat_message.py
+++ b/haystack_experimental/dataclasses/chat_message.py
@@ -127,7 +127,9 @@ class ImageContent:
         meta: Optional[Dict[str, Any]] = None,
     ) -> "ImageContent":
         """
-        Create an ImageContent object from a URL.
+        Create an ImageContent object from a URL. The image is downloaded and converted to a base64 string.
+
+        For PDF to ImageContent conversion, use the `PDFToImageContent` component.
 
         :param url:
             The URL of the image.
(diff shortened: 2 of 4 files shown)


haystack_experimental/components/image_converters/file_to_image.py
```diff
@@ -69,11 +69,11 @@ def run(
...
             List of file paths or ByteStream objects to convert.
         :param meta:
-            Optional metadata to attach to the documents.
+            Optional metadata to attach to the ImageContent objects.
             This value can be a list of dictionaries or a single dictionary.
-            If it's a single dictionary, its content is added to the metadata of all produced documents.
+            If it's a single dictionary, its content is added to the metadata of all produced ImageContent objects.
             If it's a list, its length must match the number of sources as they're zipped together.
-            For ByteStream objects, their `meta` is added to the output documents.
+            For ByteStream objects, their `meta` is added to the output ImageContent objects.
```
sjrl, line 76: Thanks for fixing these.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -67,6 +69,98 @@ def __repr__(self) -> str:
...
+        *,
+        retry_attempts: int = 2,
+        timeout: int = 10,
+        size: Optional[Tuple[int, int]] = None,
+        mime_type: Optional[str] = None,
+        detail: Optional[Literal["auto", "high", "low"]] = None,
+        meta: Optional[Dict[str, Any]] = None,
+    ) -> "ImageContent":
+        """
+        Create an ImageContent object from a URL.
```
sjrl, line 130: I'd recommend a slight rewording to make it clear we download the image from the URL, otherwise it might sound like we just store the URL in the ImageContent object.

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -67,6 +69,98 @@ def __repr__(self) -> str:
...
+            An ImageContent object.
+        """
+        # to avoid a circular import
+        from haystack_experimental.components.image_converters import ImageFileToImageContent
+
+        fetcher = LinkContentFetcher(raise_on_failure=True, retry_attempts=retry_attempts, timeout=timeout)
+        bytestream = fetcher.run(urls=[url])["streams"][0]
+
+        if mime_type:
+            bytestream.mime_type = mime_type
```
sjrl, line 159: I'm thinking we should do some additional checking here. Basically from what I can see we aren't warning the user if they provide a URL that doesn't actually point to an image. E.g. if they point to a normal website or a PDF link then there will probably a be a hard to debug error that occurs in the ImageFileToImageContent component. 

So I wonder if we should print a warning message if the inferred mime type from the LinkContentFetcher isn't an expected image format. What do you think?

haystack_experimental/dataclasses/chat_message.py
```diff
@@ -67,6 +69,98 @@ def __repr__(self) -> str:
         fields_str = ", ".join(fields)
         return f"{self.__class__.__name__}({fields_str})"
 
+    @classmethod
+    def from_file_path(
```
sjrl, line 73: Did we want to also support PDF files with this utility method?

test/dataclasses/test_chat_message.py
```diff
@@ -515,3 +516,91 @@ def test_to_openai_dict_format_invalid():
...
+
+    with patch("haystack.components.fetchers.link_content.httpx.Client.get") as mock_get:
+        mock_get.side_effect = httpx.HTTPStatusError("403 Client Error", request=Mock(), response=Mock())       
+
+        with pytest.raises(httpx.HTTPStatusError):
+            ImageContent.from_url(
+                url="https://non_existent_website_dot.com/image.jpg",
+                retry_attempts=0,
+                timeout=1,
+            )
```
sjrl, line 606: I think it would be good to add some tests for when users provide file paths or urls that point to non-image file types to see what kinds of errors occur. E.g. a .txt file, .html, etc.

haystack_experimental/dataclasses/chat_message.py
anakin87: Makes sense. Done in https://github.com/deepset-ai/haystack-experimental/pull/294/commits/c955b14364deceaf997d1ca7da8e66f16cef9332

haystack_experimental/dataclasses/chat_message.py
anakin87: To keep these convenience methods simple, I would prefer to point users to the appropriate components. WDYT?

haystack_experimental/dataclasses/chat_message.py
anakin87: I started to work on this suggestion, then I realized that the code and the behavior of the method become complex if we still want to allow the user to provide a MIME type, so I would be inclined to remove the `mime_type` parameters from these methods. As discussed in https://github.com/deepset-ai/haystack-experimental/pull/294#issue-3042393043, given a file path or URL, MIME type inference is generally reliable.
Do you agree with this idea?

sjrl: > Do you agree with this idea?

Yes! That sounds good to me.

haystack_experimental/dataclasses/chat_message.py
sjrl: It's probably fine for now, but I'd make it clear in the docs which file types these convenience methods support. Then we can see how often they get used and if there is a demand for one that supports PDFs as well.

haystack_experimental/dataclasses/chat_message.py
anakin87: - I removed the `mime_type` parameter.
- I added a check for valid image MIME types. If the MIME type does not point to an image, we now raise an error.

haystack_experimental/dataclasses/chat_message.py
anakin87: I tried to improve the docstrings about this aspect

test/dataclasses/test_chat_message.py
anakin87: added

sjrl (approved):
Looks good!

PR #297, opened 2025-05-08
feat: add Jinja2 ChatMessage extension

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack/issues/9260: I'm not yet modifying the `ChatPromptBuilder` but introducing the foundations to make it work

### Proposed Changes:
- introduce a Jinja2 ChatMessage extension
    1. The `{% message %}` tag is used to define a chat message.
    2. The message can contain text and other structured content parts.
    3. To include a structured content part in the message, the `| for_template` filter is used.
       The filter serializes the content part into a JSON string and wraps it in a `<haystack_content_part>` tag.
    4. The `_build_chat_message_json` method of the extension parses the message content parts,
       converts them into a `ChatMessage` object and serializes it to a JSON string.
    5. The obtained JSON string will be usable in the `ChatPromptBuilder` component, where templates are rendered to actual `ChatMessage` objects.

### How did you test it?
CI; several tests for the extension

### Notes for the reviewer
- Don't get scared by the size of this PR (there are many tests)
- (I'll keep working on the `ChatPromptBuilder` using this basis)

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/dataclasses/chat_message.py
+++ b/haystack_experimental/dataclasses/chat_message.py
@@ -18,6 +18,32 @@ logger = logging.getLogger(__name__)
 
 ChatMessageContentT = Union[TextContent, ToolCall, ToolCallResult, ImageContent]
 
+def _deserialize_content_part(part: Dict[str, Any]) -> ChatMessageContentT:
+    """
+    Deserialize a single content part of a serialized ChatMessage.
+
+    :param part:
+        A dictionary representing a single content part of a serialized ChatMessage.
+    :returns:
+        A ChatMessageContentT object.
+    :raises ValueError:
+        If the part is not a valid ChatMessageContentT object.
+    """
+    if "text" in part:
+        return TextContent(text=part["text"])
+    if "tool_call" in part:
+        return ToolCall(**part["tool_call"])
+    if "tool_call_result" in part:
+        result = part["tool_call_result"]["result"]
+        origin = ToolCall(**part["tool_call_result"]["origin"])
+        error = part["tool_call_result"]["error"]
+        tcr = ToolCallResult(result=result, origin=origin, error=error)
+        return tcr
+    if "image" in part:
+        return ImageContent(**part["image"])
+    raise ValueError(f"Unsupported part in serialized ChatMessage: `{part}`")
+    
+
 
 def _deserialize_content(serialized_content: List[Dict[str, Any]]) -> List[ChatMessageContentT]:
     """
@@ -29,30 +55,32 @@ def _deserialize_content(serialized_content: List[Dict[str, Any]]) -> List[ChatM
     :returns:
         Deserialized `content` field as a list of `ChatMessageContentT` objects.
     """
-    content: List[ChatMessageContentT] = []
-
-    for part in serialized_content:
-        if "text" in part:
-            content.append(TextContent(text=part["text"]))
-        elif "tool_call" in part:
-            content.append(ToolCall(**part["tool_call"]))
-        elif "tool_call_result" in part:
-            result = part["tool_call_result"]["result"]
-            origin = ToolCall(**part["tool_call_result"]["origin"])
-            error = part["tool_call_result"]["error"]
-            tcr = ToolCallResult(result=result, origin=origin, error=error)
-            content.append(tcr)
-        elif "image" in part:
-            content.append(ImageContent(**part["image"]))
-        else:
-            raise ValueError(f"Unsupported part in serialized ChatMessage: `{part}`")
-
-    return content
+    return [_deserialize_content_part(part) for part in serialized_content]
 
 
 # Note: this is a monkey patch to the original _deserialize_content function
 haystack.dataclasses.chat_message._deserialize_content = _deserialize_content
 
+def _serialize_content_part(part: ChatMessageContentT) -> Dict[str, Any]:
+    """
+    Serialize a single content part of a ChatMessage.
+
+    :param part:
+        A ChatMessageContentT object.
+    :returns:
+        A dictionary representing the content part.
+    :raises TypeError:
+        If the part is not a valid ChatMessageContentT object.
+    """
+    if isinstance(part, TextContent):
+        return {"text": part.text}
+    elif isinstance(part, ToolCall):
+        return {"tool_call": asdict(part)}
+    elif isinstance(part, ToolCallResult):
+        return {"tool_call_result": asdict(part)}
+    elif isinstance(part, ImageContent):
+        return {"image": asdict(part)}
+    raise TypeError(f"Unsupported type in ChatMessage content: `{type(part).__name__}` for `{part}`.")
 
 @dataclass
 class ChatMessage(HaystackChatMessage):
@@ -116,6 +144,7 @@ class ChatMessage(HaystackChatMessage):
                 raise ValueError("The user message must contain at least one textual part.")
 
         return cls(_role=ChatRole.USER, _content=content, _meta=meta or {}, _name=name)
+    
 
     def to_dict(self) -> Dict[str, Any]:
         """
@@ -129,20 +158,8 @@ class ChatMessage(HaystackChatMessage):
         serialized["role"] = self._role.value
         serialized["meta"] = self._meta
         serialized["name"] = self._name
-        content: List[Dict[str, Any]] = []
-        for part in self._content:
-            if isinstance(part, TextContent):
-                content.append({"text": part.text})
-            elif isinstance(part, ToolCall):
-                content.append({"tool_call": asdict(part)})
-            elif isinstance(part, ToolCallResult):
-                content.append({"tool_call_result": asdict(part)})
-            elif isinstance(part, ImageContent):
-                content.append({"image": asdict(part)})
-            else:
-                raise TypeError(f"Unsupported type in ChatMessage content: `{type(part).__name__}` for `{part}`.")
 
-        serialized["content"] = content
+        serialized["content"] = [_serialize_content_part(part) for part in self._content]
         return serialized
 
     def to_openai_dict_format(self) -> Dict[str, Any]:
--- a/haystack_experimental/dataclasses/chat_message.py
+++ b/haystack_experimental/dataclasses/chat_message.py
@@ -42,7 +42,7 @@ def _deserialize_content_part(part: Dict[str, Any]) -> ChatMessageContentT:
     if "image" in part:
         return ImageContent(**part["image"])
     raise ValueError(f"Unsupported part in serialized ChatMessage: `{part}`")
-    
+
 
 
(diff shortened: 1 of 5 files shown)


haystack_experimental/dataclasses/chat_message.py
```diff
@@ -18,6 +18,32 @@
 
 ChatMessageContentT = Union[TextContent, ToolCall, ToolCallResult, ImageContent]
 
+def _deserialize_content_part(part: Dict[str, Any]) -> ChatMessageContentT:
```
anakin87, line 21: `_deserialize_content_part ` and `_serialize_content_part` have been extracted to be used in the Jinja2 extension.

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
```
sjrl, line 1: I think we are missing an `__init__.py` file in this folder

test/utils/test_jinja_chat_extension.py
```diff
@@ -0,0 +1,339 @@
...
+        <haystack_content_part>{"this is": "invalid" json}</haystack_content_part>
+        <haystack_content_part>not even trying to be json</haystack_content_part>
+        <haystack_content_part>{]</haystack_content_part>
+        Normal text after.
+        {% endmessage %}
+        """
+        with pytest.raises(json.JSONDecodeError):
+            jinja_env.from_string(template).render()
+
+
```
sjrl, line 339: Could we also add some tests for when a user forgets a `{% endmessage %}` tag? I'd also be curious to know what happens in this case and what the error message would be.

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    A Jinja2 extension for creating structured chat messages with mixed content types.
+
+    This extension provides a custom {% message %} tag that allows creating chat messages
+    with different attributes (role, name, meta) and mixed content types (text, images, etc.).
+
+    Inspired by [Banks](https://github.com/masci/banks).
+
+    Example:
+    {% message role="system" %}
```
sjrl, line 36: Is the role always required when using the `message` tag?

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
...
+    {% for image in images %}
+    {{ image | for_template }}
+    {% endfor %}
+    {% endmessage %}
+
+    ### How it works
+    1. The {% message %} tag is used to define a chat message.
+    2. The message can contain text and other structured content parts.
+    3. To include a structured content part in the message, the `| for_template` filter is used.
+       The filter serializes the content part into a JSON string and wraps it in a `<haystack_content_part>` tag.
```
sjrl, line 51: Does this mean manual insertion of `<haystack_content_part>` isn't expected of users to do? Or would it be a good option for advanced users?

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
...
+        :param meta: Optional metadata dictionary
+        :param caller: Callable that returns the rendered content
+        :return: A JSON string representation of the ChatMessage object
+        :raises ValueError: If the message content is empty
+        """
+
+        content = caller()
+        parts = self._parse_content_parts(content)
+        if not parts:
+            raise ValueError("Message content is empty")
```
sjrl, line 135: Would it be possible to provide some additional contextual information here? Like print out a snippet of the incorrect Jinja2 template and/or a line number? I'm just imagining for long templates or when you have multiple ChatPromptBuilders in a pipeline it could be hard to go find the wrongly formatted template based on this error message alone.

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
...
+        message = ChatMessage(_role=ChatRole(role), _content=parts, _name=name, _meta=meta)
```
sjrl, line 136: Just a curiosity thing, why not use the `from_role` methods for this?

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
...
+    {% endmessage %}
+
+    ### How it works
+    1. The {% message %} tag is used to define a chat message.
+    2. The message can contain text and other structured content parts.
+    3. To include a structured content part in the message, the `| for_template` filter is used.
+       The filter serializes the content part into a JSON string and wraps it in a `<haystack_content_part>` tag.
+    4. The `_build_chat_message_json` method of the extension parses the message content parts,
+       converts them into a ChatMessage object and serializes it to a JSON string.
+    5. The obtained JSON string is usable in the ChatPromptBuilder component, where templates are rendered to actual
```
sjrl, line 54: I'm interested to see step 5! Since it feels a little surprising to me that you need to make a ChatMessage then JSONify it and then recreate it again in ChatPromptBuilder. But I'm sure there is a reason for it :D

test/utils/test_jinja_chat_extension.py
```diff
@@ -0,0 +1,339 @@
...
+    @pytest.fixture
+    def jinja_env(self) -> SandboxedEnvironment:
+        # we use a SandboxedEnvironment here to replicate the conditions of the ChatPromptBuilder component
+        env = SandboxedEnvironment(extensions=[ChatMessageExtension])
+        env.filters["for_template"] = for_template
+        return env
+
+    def test_message_with_name_and_meta(self, jinja_env):
+        template = """
+        {% message role="user" name="Bob" meta={"language": "en"} %}
```
sjrl, line 25: Could we also add tests checking for the errors that occur when name and/or meta are malformed? E.g. if meta isn't valid JSON

haystack_experimental/utils/jinja_chat_extension.py
```diff
@@ -0,0 +1,203 @@
...
+    Hello! I am {{user_name}}. Please describe the images.
+    {% for image in images %}
+    {{ image | for_template }}
+    {% endfor %}
+    {% endmessage %}
+
+    ### How it works
+    1. The {% message %} tag is used to define a chat message.
+    2. The message can contain text and other structured content parts.
+    3. To include a structured content part in the message, the `| for_template` filter is used.
```
sjrl, line 50: Also more of just a question. Why the name `for_template`? I'm not yet quite understanding what it does based on just the name.

sjrl:
@anakin87 Overall looks really good! I have some some comments mostly surrounding tests and error handling but the templating language itself looks very solid!

test/utils/test_jinja_chat_extension.py
anakin87: `jinja2.exceptions.TemplateSyntaxError: Unexpected end of template. Jinja was looking for the following tags: 'endmessage'. The innermost block that needs to be closed is 'message'.`

Test added in https://github.com/deepset-ai/haystack-experimental/pull/297/commits/e123c9b544dcbb99469f670412eabf7d082827bb

haystack_experimental/utils/jinja_chat_extension.py
anakin87: Yes.
I would add an empty `__init__.py` file because I don't want the ChatMessage extension to be easily accessible/usable outside the `ChatPromptBuilder`

haystack_experimental/utils/jinja_chat_extension.py
anakin87: Yes. Added a test in https://github.com/deepset-ai/haystack-experimental/pull/297/commits/2a28e335dd0bb350a8ec196d8e85acec538d0dec.

I think it is reasonable. (We might also decide to use `user` if the role is not provided but I'm not sure...)

haystack_experimental/utils/jinja_chat_extension.py
anakin87: While this is technically feasible, I would say that if we design our abstraction well, I hope that there is no need for users to know the underlying mechanism and interfere with it. It would just be an implementation detail that we can improve/alter in the future (without worrying about anyone being impacted).

haystack_experimental/utils/jinja_chat_extension.py
sjrl: I think it's fine to require. We can revisit later if it becomes a sticking point for users.

haystack_experimental/utils/jinja_chat_extension.py
anakin87: I've tried to improve the checks for empty content but now most errors are triggered because the content is empty or only consists of whitespace chars. I cannot trigger an error for non-empty content producing 0 content parts.
https://github.com/deepset-ai/haystack-experimental/pull/297/commits/ebf084d482cc0f393ad190c66e1c8c678f5d4903

haystack_experimental/utils/jinja_chat_extension.py
anakin87: I agree that it is a bit tricky. The main reason is that Jinja2 mostly work with strings: this is especially true for the [`SandboxedEnvironment`](https://jinja.palletsprojects.com/en/stable/sandbox/) that we use for security reasons.

This forces us to use JSON strings as an intermediate format.

During my investigation for POC, I found that bypassing this limitation would require using `NativeEnvironment`, which introduces security risks.

sjrl: Okay no worries then! Thanks for the explanation.

haystack_experimental/utils/jinja_chat_extension.py
anakin87: Good question!
Initially I did this for simplicity but then I realized that I was not validating messages: we usually don't need to do this because the preferred way to build messages is via class methods like `ChatMessage.from_user(),` etc. that force users to align with a specific signature.

For this reason, even if not particularly nice, I've added a validation method + tests in https://github.com/deepset-ai/haystack-experimental/pull/297/commits/736c694891b47c05fa96071020cc035fba665b93

haystack_experimental/utils/jinja_chat_extension.py
anakin87: Good point. I would like a short but descriptive name.
Other ideas: `content_part`/`prepare_content_part`/`mark_content_part`/`format_content_part`; something related to "including".

Do you have any other ideas?
(We can also involve some other team members if necessary)

sjrl: Yeah I'm not entirely sure either. I'd be happy to have other team members give their input

sjrl (approved):
Everything looks good! I don't think the `for_template` naming needs to block this. Will leave it up to you

anakin87:
I now renamed the filter to `templatize_part`, as we decided.

PR #299, opened 2025-05-12
feat: extent `ChatPromptBuilder` to support string templates

anakin87:
### Related Issues

- fixes #issue-number

### Checklist

- I documented my code


haystack_experimental/components/builders/chat_prompt_builder.py
```diff
@@ -0,0 +1,341 @@
...
+            self._env.add_extension(Jinja2TimeExtension)
+
+        extracted_variables = []
+        if template and not variables:
+            if isinstance(template, list):
+                for message in template:
+                    if message.is_from(ChatRole.USER) or message.is_from(ChatRole.SYSTEM):
+                        # infer variables from template
+                        if message.text is None:
+                            raise ValueError(f"The provided ChatMessage has no text. ChatMessage: {message}")
```
sjrl, line 168: Let's add a bit more of a call to action here. So say that text always need to be added to the ChatMessage

haystack_experimental/components/builders/chat_prompt_builder.py
```diff
@@ -0,0 +1,341 @@
...
+
+        if isinstance(template, list) and not all(isinstance(message, ChatMessage) for message in template):
+            raise ValueError(
+                f"The {self.__class__.__name__} expects a list containing only ChatMessage instances. "
+                f"The provided list contains other types. Please ensure that all elements in the list "
+                f"are ChatMessage instances."
+            )
+
+        processed_messages = []
+        if isinstance(template, list):
```
sjrl, line 244: Would it be straightforward to add a warning if a user passes a list of ChatMessages and tries to use the `templatize_part` jinja filter? Basically to say that this is not allowed?

test/components/builders/test_chat_prompt_builder.py
```diff
@@ -0,0 +1,930 @@
...
+        }
+        
+    def test_from_dict(self):
+        template = """
+        {% message role="user" %}
+        Hello, my name is {{name}}!
+        {% endmessage %}
+
+        {% message role="assistant" %}
+        Hello, I am {{assistant_name}}! How can I help you today?
```
sjrl, line 915: Are we missing the endmessage tag here?

test/components/builders/test_chat_prompt_builder.py
```diff
@@ -0,0 +1,930 @@
...
+        {% endmessage %}
+        """
+        with pytest.raises(TemplateSyntaxError):
+            ChatPromptBuilder(template=template)        
+
+    def test_run(self):
+        template = """
+        {% message role="user" %}
+        Hello, my name is {{name}}!
+        {% endmessage %}
```
sjrl, line 708: Could we also add a multiline example to double check we preserve newline characters for long messages e.g. 
```python
        {% message role="user" %}
        Hello, my name is {{name}}!
        Second line.
        Third line.
        {% endmessage %}
```

test/components/builders/test_chat_prompt_builder.py
sjrl: And if so should this throw an error?

sjrl (approved):
Only a few minor comments. Looks really good!

haystack_experimental/components/builders/chat_prompt_builder.py
anakin87: I improved the error message. BTW, this is quite unlikely to happen if users use `ChatMessage.from_user` etc.. It was added to make mypy happy.

haystack_experimental/components/builders/chat_prompt_builder.py
anakin87: done in https://github.com/deepset-ai/haystack-experimental/pull/299/commits/7bca60c3ceb26891a3341613cef30c318a08bff4

test/components/builders/test_chat_prompt_builder.py
anakin87: good catch!

test/components/builders/test_chat_prompt_builder.py
anakin87: added in https://github.com/deepset-ai/haystack-experimental/pull/299/commits/d86bd2692167551ea921b7a6eacf0dcd3c3f280c

PR #305, opened 2025-05-15
feat: `ImageContent` validation

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9363

### Proposed Changes:
- Add `ImageContent` validation in `__post_init__`
  - check if `base64_image` is a valid base64 string
  - if `mime_type` is not `None` (because provided or inferred), check if it corresponds to an image
- other minor refactoring

### How did you test it?
CI; new tests; updated tests

### Checklist

- I documented my code


haystack_experimental/dataclasses/image_content.py
```diff
@@ -43,23 +75,25 @@ class ImageContent:
     meta: Dict[str, Any] = field(default_factory=dict)
 
     def __post_init__(self):
+        try:
+            decoded_image = base64.b64decode(self.base64_image, validate=True)
+        except Exception as e:
+            raise ValueError("The base64 string is not valid") from e
```
sjrl, line 81: Could you provide some timings on how fast/slow this is for large images (e.g. 10-20 mb)?

If this is always fast for even large images I'm happy to do this otherwise I'm worried it would negatively impact run times. Previously I wasn't too worried since we only did this if the mimetype wasn't provided so there was a way to avoid this potentially slow step.

haystack_experimental/dataclasses/image_content.py
```diff
@@ -176,6 +210,10 @@ def from_url(
             msg = f"The URL does not point to an image. The MIME type of the URL is {bytestream.mime_type}."
             raise ValueError(msg)
 
+        if bytestream.mime_type == "application/pdf":
+            raise ValueError("PDF files are not supported. "
+                             "For PDF to ImageContent conversion, use the `PDFToImageContent` component.")
+
```
sjrl, line 216: good catch!

sjrl:
@anakin87 looks good! Only one comment on the how fast the decoding step is for large images

haystack_experimental/dataclasses/image_content.py
anakin87: I measured decoding times in this [Colab notebook](https://colab.research.google.com/drive/1anCslM2b1wh6qX5GkPE08xOw5cHwymGV) for a ~20 Mb image.
Decoding takes ~130 ms in Colab.

I did this experiment on Colab because it only has 2 CPU cores and it's not a very powerful environment.
On my Mac, decoding takes ~20 ms.

I would say that these times are acceptable, given that decoding with validation helps to prevent downstream errors. WDYT?

sjrl: Hmm I'm not entirely sure. Could we add a boolean to toggle whether we would like to validate? Then if users really care about performance they could skip this step.

anakin87: This might a be premature optimization but I implemented it in https://github.com/deepset-ai/haystack-experimental/pull/305/commits/08e45f72144d3cde12360d2a5b18f85d271f1801

Anyway, I would like to keep things simple: either we do the whole process or we skip it.

sjrl (approved):
Looks good!

PR #311, opened 2025-06-02
feat: Add `DocumentToImageConent` component to help enable RAG with image Documents

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9263
- partially addresses https://github.com/deepset-ai/haystack/issues/9321

### Proposed Changes:

 
 
Add `DocumentToImageContent` component and provide an example of how to query a RAG pipeline with Image and Text-Only Documents.

**DocumentToImage Component Description**

This component processes a list of Documents and extracts visual content from supported file formats, converting
them into ImageContents that can be used for multimodal AI tasks. It handles both direct image files and PDF
documents by extracting specific pages as images.

The component categorizes input documents into three groups:
- Image documents: Successfully converted to ImageContent objects
- Non-image documents: Documents that don't contain supported image formats
- Documents with missing information: Documents lacking required metadata such as `file_path` or `page_number`.

Documents will be skipped and categorized as non-image documents in the following scenarios:
- The `file_path` is not present in the document metadata
- The file path does not point to an existing file when combined with `root_path`
- The file format is not among the supported image types
- For PDF files, the `page_number` is not present in the metadata

Example:
```python
converter = DocumentToImageContent(
    root_path="/data/documents",
    detail="high",
    size=(800, 600)
)

documents = [
    Document(content="Optional description of image.jpg", meta={"file_path": "image.jpg"}),
    Document(content="Text content of page 1 of doc.pdf", meta={"file_path": "doc.pdf", "page_number": 1})
]

result = converter.run(documents)
image_documents = result["image_documents"]
image_contents = result["image_contents"]
```

### How did you test it?

- Added unit tests
- Using the example script in the PR

### Notes for the reviewer

I provided an example script to see how this component would be used in a pipeline. When this PR is ready for review I will take it out and move it into the cookbook. 

### Checklist

- I documented my code


sjrl:
@davidsbatista I've removed you and added @anakin87 as the reviewer since he has more context on the multimodality feature

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+        ```
+    """
+
+    def __init__(
+        self,
+        *,
+        root_path: Optional[str] = None,
+        detail: Optional[Literal["auto", "high", "low"]] = None,
+        size: Optional[Tuple[int, int]] = None
+    ):
```
anakin87, line 64: does it make sense to introduce sth like `file_path_meta_field` to make the source of the image configurable?

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+            logger.warning(
+                "In DocumentToImageContent {len_missing_file_paths} Documents are either missing a `file_path` "
+                "key in their metadata or the `file_path` does not point to a valid file. "
+                "They will be returned in the `non_image_documents` output.",
+                len_missing_file_paths=len(missing_info_docs),
+            )
+
+        # We de-duplicate the pdf documents because it's possible that the same PDF page is represented by multiple
+        # documents
+        pdf_docs = self._deduplicate(pdf_docs)
```
anakin87, line 172: 1. I imagine that deduplication is performed to avoid creating multiple duplicate `ImageContent` objects. Right?
2. What happens if the text of a PDF page is split into N chunks? Is it possible that we lose some textual information in this case? I would like to avoid it. **EDIT**: I found a test that covers this exact use case. WDYT? As a user, I would find it unexpected, but maybe there are good reasons...

haystack_experimental/components/image_converters/document_to_image.py
sjrl: yup! I like the idea from your `ImageEmbedder` component. I'll do that here too

haystack_experimental/components/image_converters/document_to_image.py
anakin87: OK with avoiding the deduplication for now.
We can re-evaluate this in the future if there is interest for this feature. (Unless you think it's a must-have).

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+        root_path: Optional[str] = None,
+        detail: Optional[Literal["auto", "high", "low"]] = None,
+        size: Optional[Tuple[int, int]] = None
+    ):
+        """
+        Initialize the DocumentToImageContent component.
+
+        :param root_path: The root directory path where document files are located. If provided, file paths in
+            document metadata will be resolved relative to this path. If None, file paths are treated as absolute paths.
+        :param detail: Optional detail level of the image (only supported by OpenAI). One of "auto", "high", or "low".
```
dfokina, line 70: What does this do exactly, maybe a link to openai docs on this would help?
```suggestion
        :param detail: Optional detail level of the image (only supported by OpenAI). Can be "auto", "high", or "low".
```

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+            This will be passed to the created ImageContent objects.
+        :param size: If provided, resizes the image to fit within the specified dimensions (width, height) while
```
dfokina, line 72:
```suggestion
        :param size: If provided, resizes the image to fit specified dimensions (width, height) while
```

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+        Deduplicate the documents based on file path and page number if available.
+
+        This deduplication is particularly important for PDF documents where the same page might be represented by
+        multiple Documents due to text splitting or other preprocessing steps. Only the first occurrence
+        of each unique (file_path, page_number) combination is retained.
+
+        :param documents: List of Documents to deduplicate.
+
+        :returns:
+            List of Documents with duplicates removed, maintaining original order of first occurrences.
```
dfokina, line 96:
```suggestion
        multiple documents due to text splitting or other preprocessing steps. Only the first occurrence
        of each unique (file_path, page_number) combination is retained.

        :param documents: A list of documents to deduplicate.

        :returns:
            A list of documents with duplicates removed, maintaining original order of first occurrences.
```

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+                unique_documents.append(doc)
+                seen.add(key)
+        return unique_documents
+
+    @component.output_types(
+        image_documents=List[Document], image_contents=List[ImageContent], non_image_documents=List[Document]
+    )
+    def run(self, documents: List[Document]) -> Dict[str, List[Document]]:
+        """
+        Convert documents with image or PDF sources to ImageContents.
```
dfokina, line 112:
```suggestion
        Convert documents with image or PDF sources into ImageContent objects.
```

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,204 @@
...
+
+        This method processes the input documents, extracting images from supported file formats and converting them
+        to ImageContents. Documents are categorized based on their file type and processing success.
+
+        :param documents: List of Documents to process. Each document should have metadata containing at minimum
```
dfokina, line 117:
```suggestion
        :param documents: A list of documents to process. Each document should have metadata containing at minimum
```

sjrl:
@anakin87 I've opted for raising errors rather than warnings especially considering recent requests that users can often prefer errors being raised than "failing-forward" with warnings. My only thought could be that we could add a `raise_on_failure`/`raise_on_error` flag in the init which would be set to True by default. Then users could control this behavior. 

But happy to add this later if it's requested rather than pre-emptively doing so.

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,178 @@
...
+                    f" Please ensure that the documents you are trying to convert have this key set."
+                )
+
+            if not Path(self.root_path, file_path).is_file():
+                raise ValueError(
+                    f"Document with ID '{doc.id}' has an invalid file path '{file_path}'. "
+                    f"Please ensure that the documents you are trying to convert have valid file paths."
+                )
+
+            # Store the non-image documents separately
```
anakin87, line 135:
```suggestion
```

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,178 @@
...
+        self.file_path_meta_field = file_path_meta_field
+        self.root_path = root_path or ""
+        self.detail = detail
+        self.size = size
+
+        # Initializing these converters will trigger the lazy import checks.
+        self._file_to_image_converter = ImageFileToImageContent(detail=detail, size=size)
+        self._pdf_to_image_converter = PDFToImageContent(detail=detail, size=size)
+
+    @component.output_types(image_documents=List[Document], image_contents=List[ImageContent])
```
anakin87, line 91: I have the impression that now we are returning the original list of Documents, unless we encounter errors during the process. In this case, we might simplify this component to only return a list of `ImageContent`.
Makes sense or am I missing some details?

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,180 @@
...
+    def run(
+        self, documents: List[Document]
+    ) -> Union[Dict[str, List[Document]], Dict[str, List[ImageContent]], Dict[str, List]]:
```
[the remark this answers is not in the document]
anakin87: could you please explain?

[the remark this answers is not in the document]
anakin87: could you please explain?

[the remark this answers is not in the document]
sjrl: It has to do with this line 
```python
        if not documents:
            return {"image_documents": [], "image_contents": []}
```
mypy complains that this return is not satisified with just `Union[Dict[str, List[Document]], Dict[str, List[ImageContent]]]`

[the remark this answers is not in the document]
anakin87: thanks

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,180 @@
...
+        into ImageContent objects. Documents are categorized based on their file type and processing success.
+
+        :param documents: A list of documents to process. Each document should have metadata containing at minimum
+            a 'file_path_meta_field' key. PDF documents additionally require a 'page_number' key to specify which
+            page to convert.
+
+        :returns:
+            Dictionary containing one key:
+            - "image_contents": ImageContents created from the processed documents. These contain base64-encoded image
+                data and metadata. The order corresponds to order of input documents.
```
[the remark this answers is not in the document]
sjrl: yeah preserving order is probably a good idea. I'll work on that

haystack_experimental/components/image_converters/document_to_image.py
```diff
@@ -0,0 +1,180 @@
...
+    This component processes a list of documents and extracts visual content from supported file formats, converting
+    them into ImageContents that can be used for multimodal AI tasks. It handles both direct image files and PDF
+    documents by extracting specific pages as images.
+
+    Documents are expected to have metadata containing:
+    - The `file_path_meta_field` is not present in the document metadata
+    - The file path does not point to an existing file when combined with `root_path`
+    - The file format is not among the supported image types
+    - For PDF files, the `page_number` is not present in the metadata
+    Otherwise, the component will raise a `ValueError` with a descriptive message.
```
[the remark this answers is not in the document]
sjrl: rewritten

PR #314, opened 2025-06-04
ImageEmbedder POC

anakin87:
### Related Issues

- fixes #issue-number

### Checklist

- I documented my code


haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,215 @@
...
+                "In case you want to embed a list of strings, please use the SentenceTransformersTextEmbedder."
+            )
+        if self.embedding_backend is None:
+            raise RuntimeError("The embedding model has not been loaded. Please call warm_up() before running.")
+
+        images_to_embed = []
+        for doc in documents:
+            image_path = doc.meta.get(self.meta_field_for_image_path)
+            if not image_path:
+                raise ValueError(f"Image path not found in document metadata for document {doc.id}")
```
sjrl, line 200: Yeah I wondered about this type of logic. You can see I went for a slightly different approach in `DocumentToImageContent`. 

In general I think we should think where in the indexing pipeline do we expect users to split documents into different streams (e.g. Image Docs, PDF Docs, and Non-Image Docs). For example, I chose for `DocumentToImageContent` to support all three types instead of raising errors. 

What is your take? Should we use something like a FileTypeRouter or a ConditionalRouter earlier in the Indexing pipeline and enforce that at this stage we only expect Image Documents in the Image Embedders?

haystack_experimental/components/embedders/cohere_doc_image_embedder.py
```diff
@@ -0,0 +1,147 @@
...
+            ## let's take first non None tuple as that's the one we want
+            for emb_tuple in response.embeddings:
+                # emb_tuple[0] is a str denoting the embedding type (e.g. "float", "int8", etc.)
+                if emb_tuple[1] is not None:
+                    # ok we have embeddings for this type, let's take all the embeddings (a list of embeddings)
+                    all_embeddings.extend(emb_tuple[1])
+
+        for doc, embedding in zip(documents, all_embeddings):
+            doc.embedding = embedding
+            doc.meta["embedding_source"] = f"meta.{self.meta_field_for_image_path}"
```
sjrl, line 145: Out of curiosity why add this `embedding_source` field?

haystack_experimental/components/embedders/cohere_doc_image_embedder.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    A component for computing Document embeddings based on images.
+
+    The embedding of each Document is stored in the `embedding` field of the Document.
+    """
+
+    def __init__(
+        self,
+        *,
+        meta_field_for_image_path: str = "image_path",
```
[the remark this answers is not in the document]
anakin87: OK, I now understand better looking at your PR. We expect `DocumentToImageContent` to work with both manually created Documents containing a reference to an image path and regular textual Documents created by our converters which may contain a path to an image (PDF) or not.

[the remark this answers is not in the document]
sjrl: Also I imagine we will create a new component called something like `ImageContentToDocument` which can take in an `ImageContent` object either run OCR, image captioning, etc. and convert it into a Document. Then we'd pass this off to an embedder like `ImageEmbedder` or our normal `DocumentEmbedder` and then to the DocumentWriter.

haystack_experimental/components/embedders/cohere_doc_image_embedder.py
sjrl: ahh okay I see. I think we might want to make it called `embedding_type` and give it values like `image` vs. `text`. I think `embedding_source` just point to the file path doesn't given enough info to know what was used to make the embedding.

anakin87: Good idea.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
sjrl: Thinking more on this, I think the `FileTypeRouter` is the right component for Indexing to separate the different files. We may want to make something like a `PDFRouter` as well, but that can come later. 

At query time (so this pertains more to my PR for `DocumentToImageContent`) I think we should make a `DocumentTypeRouter` which works like the `FileTypeRouter` except it looks at the `file_path` in the meta field to determine the mimetype. I think this is the most extensible approach when we add more modalities like audio, video, etc. Then I'd have `DocumentToImageContent` raise an error if it finds an unsupported file type similar to `ImageEmbedder`.

anakin87: If you think it's better, I would be OK with this approach.

anakin87:
Closing in favor of a proper implementation in #319

PR #317, opened 2025-06-05
fix: refactor code and update `init_params` in `debug_state`

Amnah199:
### Related Issues

- fixes [#9464](https://github.com/deepset-ai/haystack/issues/9464)
- remove the methods that are not needed any longer

### Proposed Changes:

- remove `remove_unserializable_data` method (not needed after [this PR](https://github.com/deepset-ai/haystack/pull/9497) is merged)
- Reintroduce params in `debug_state` that are set during initialization. This was removed for simplicity while converging with haystack/main.

### How did you test it?

Ran the tests
### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/core/pipeline/pipeline.py
+++ b/haystack_experimental/core/pipeline/pipeline.py
@@ -3,16 +3,17 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import json
-from copy import deepcopy
+from copy import copy, deepcopy
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Mapping, Optional, Set, Tuple, Union, cast
 
 from haystack import Answer, Document, ExtractedAnswer, logging, tracing
-from haystack.components.joiners import BranchJoiner, DocumentJoiner
+from haystack.components.joiners import DocumentJoiner
 from haystack.core.component import Component
 from haystack.dataclasses import ChatMessage, GeneratedAnswer, SparseEmbedding
 from haystack.telemetry import pipeline_running
+from haystack.utils import _serialize_value_with_schema
 
 from haystack_experimental.core.errors import (
     PipelineBreakpointException,
@@ -289,10 +290,21 @@ class Pipeline(PipelineBase):
                             component_inputs[key] = deserialize_component_input(value)
 
                     if validated_breakpoints and not self.resume_state:
-                        state_inputs_serialised = remove_unserializable_data(deepcopy(inputs))
+                        state_inputs_serialised = deepcopy(inputs)
                         # inject the component_inputs into the state_inputs so we can this component init params in
                         # the JSON state
-                        state_inputs_serialised[component_name] = remove_unserializable_data(deepcopy(component_inputs))
+                        state_inputs_serialised[component_name] = deepcopy(component_inputs)
+
+                        # TODO: Adding the old approach until we finalize if we
+                        # want to use serialization with schema or without
+
+                        # We use copy instead of deepcopy to avoid issues with unpickleable objects like RLock
+                        params = copy(component["instance"].__dict__)
+
+                        # this is needed as the template param is stored as _template_string in the component's __dict__
+                        # if "_template_string" in params:
+                        #    params["template"] = params["_template_string"]
+                        state_inputs_serialised[component_name]["init_parameters"] = params
 
                         Pipeline._check_breakpoints(
                             breakpoints=validated_breakpoints,
@@ -301,22 +313,13 @@ class Pipeline(PipelineBase):
                             inputs=state_inputs_serialised,
                             debug_path=self.debug_path,
                             original_input_data=data,
-                            ordered_component_names=self.ordered_component_names
+                            ordered_component_names=self.ordered_component_names,
                         )
-
-                    # the _consume_component_inputs() when applied to the DocumentJoiner inputs wraps 'documents' in an
-                    # extra list, so there's a 3 level deep list, we need to flatten it to 2 levels only
-                    instance: Component = component["instance"]
-                    if self.resume_state and isinstance(instance, DocumentJoiner):  # noqa: SIM102
-                        if isinstance(component_inputs["documents"], list):  # noqa: SIM102
-                            if isinstance(component_inputs["documents"][0], list):  # noqa: SIM102
-                                if isinstance(component_inputs["documents"][0][0], list):  # noqa: SIM102
-                                    component_inputs["documents"] = component_inputs["documents"][0]
 
                     component_outputs = self._run_component(
                         component_name=component_name,
                         component=component,
-                        inputs=component_inputs, # the inputs to the current component
+                        inputs=component_inputs,  # the inputs to the current component
                         component_visits=component_visits,
                         parent_span=span,
                     )
@@ -520,6 +523,15 @@ class Pipeline(PipelineBase):
             The saved state dictionary
         """
         dt = datetime.now()
+
+        # we store a component init_parameters together with the breakpoint in the saved state
+        # this is helpful for debugging or manually updating the state
+        for value in inputs.values():
+            if "init_parameters" in value.keys():
+                init_params = value.pop("init_parameters")
+                for k, v in value.items():
+                    if k in init_params.keys() and v is None:
+                        value[k] = serialize_component_input(init_params[k])
         state = {
             "input_data": serialize_component_input(original_input_data),  # original input data
             "timestamp": dt.isoformat(),
@@ -563,7 +575,7 @@ class Pipeline(PipelineBase):
         inputs: Dict[str, Any],
         debug_path: Optional[Union[str, Path]] = None,
         original_input_data: Optional[Dict[str, Any]] = None,
-        ordered_component_names: Optional[List[str]] = None
+        ordered_component_names: Optional[List[str]] = None,
     ):
         """
         Check if the `component_name` is in the breakpoints and if it should break.
@@ -588,10 +600,9 @@ class Pipeline(PipelineBase):
                     component_visits=component_visits,
                     debug_path=debug_path,
                     original_input_data=original_input_data,
-                    ordered_component_names=ordered_component_names
+                    ordered_component_names=ordered_component_names,
                 )
                 raise PipelineBreakpointException(msg, component=component_name, state=state)
-
 
 
 def deserialize_component_input(value):  # noqa: PLR0911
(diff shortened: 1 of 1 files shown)


haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -306,15 +312,13 @@ def run(  # noqa: PLR0915, PLR0912
...
-                    if self.resume_state and isinstance(instance, DocumentJoiner):  # noqa: SIM102
-                        if isinstance(component_inputs["documents"], list):  # noqa: SIM102
-                            if isinstance(component_inputs["documents"][0], list):  # noqa: SIM102
-                                if isinstance(component_inputs["documents"][0][0], list):  # noqa: SIM102
-                                    component_inputs["documents"] = component_inputs["documents"][0]
+
+                    if self.resume_state and self.resume_state["breakpoint"]["component"] == component_name:
+                        for socket_name, socket in component["input_sockets"].items():
+                            if is_socket_lazy_variadic(socket):
+                                component_inputs[socket_name] = component_inputs[socket_name][0]
```
Amnah199, line 322: This a replacement of the initial workaround for `DocumentJoiner` but it works only if consume_input creates a nested 3 level nested array for all lazy variadic components.

sjrl: Is this generally true? 

> but it works only if consume_input creates a nested 3 level nested array for all lazy variadic components.

Do you know why `consume_input` creates this deeply nested of an array for lazy variadic components?

haystack_experimental/components/builders/answer_builder.py
```diff
@@ -0,0 +1,200 @@
...
+from haystack import Document, component, logging
+from haystack.dataclasses.chat_message import ChatMessage
+
+from haystack_experimental.dataclasses import GeneratedAnswer
+
+logger = logging.getLogger(__name__)
+
+
+@component
+class AnswerBuilder:
```
sjrl, line 17: Sorry if I missed why, but could you briefly explain why AnswerBuilder needs to be included in Experimental?

Amnah199:
Yes the new `to_dict` method in `GeneratedAnswer` serializes ChatMessage properly, and prevents JSON serializable error. I added the answer dataclasses and made `AnswerBuilder` here to prevent failing test cases but mismatch still exists.

haystack_experimental/core/pipeline/pipeline.py
Amnah199: I looked into the code and [here](https://github.com/deepset-ai/haystack/blob/f8155e1b77db8e5d9ae48d181bd9c681e6f18c51/haystack/core/pipeline/base.py#L1081) when the socket_inputs are extracted from the socket they are added to another list. Then for `lazy_variadic` we return this nested array as `consumed_inputs`.

Amnah199: I tried to investigate how this works error free in the pipeline, but it required me more time to deep dive into this.

sjrl:
> Yes the new `to_dict` method in `GeneratedAnswer` serializes ChatMessage properly, and prevents JSON serializable error. I added the answer dataclasses and made `AnswerBuilder` here to prevent failing test cases but mismatch still exists.

Okay I gotcha. Would it be possible then to only bring over GeneratedAnswer and not all of the other answer data classes if they aren’t needed? Also it’d be great if you could inherit from the haystack GeneratedAnswer and only add the code you have to. You can see how we did this for ChatMessage in this repo as well.

Amnah199:
> > Yes the new `to_dict` method in `GeneratedAnswer` serializes ChatMessage properly, and prevents JSON serializable error. I added the answer dataclasses and made `AnswerBuilder` here to prevent failing test cases but mismatch still exists.
> 
> Okay I gotcha. Would it be possible then to only bring over GeneratedAnswer and not all of the other answer data classes if they aren’t needed? Also it’d be great if you could inherit from the haystack GeneratedAnswer and only add the code you have to. You can see how we did this for ChatMessage in this repo as well.

Right, I can do that. I initially added it as-is since I didn’t want to spend too much time patching GeneratedAnswer, given that it's going to be merged eventually. Just a heads-up — I might also need to add the AnswerJoiner to avoid the failing tests caused by mismatches.

haystack_experimental/components/builders/answer_builder.py
```diff
@@ -0,0 +1,130 @@
...
+    ```python
+    from haystack.components.builders import AnswerBuilder
+
+    builder = AnswerBuilder(pattern="Answer: (.*)")
+    builder.run(query="What's the answer?", replies=["This is an argument. Answer: This is the answer."])
+    ```
+    """
+
+    @component.output_types(answers=List[GeneratedAnswer])
+    def run(  # pylint: disable=too-many-positional-arguments
```
sjrl, line 37: Just to double check we still have this method here just to update this line 
`@component.output_types(answers=List[GeneratedAnswer])` to use the new `GeneratedAnswer` from haystack experimental right?

haystack_experimental/core/pipeline/base.py
```diff
@@ -988,7 +990,11 @@ def _consume_component_inputs(component_name: str, component: Dict, inputs: Dict
         for socket_name, socket in component["input_sockets"].items():
             socket_inputs = component_inputs.get(socket_name, [])
             socket_inputs = [sock["value"] for sock in socket_inputs if sock["value"] is not _NO_OUTPUT_PRODUCED]
-            if socket_inputs:
+
+            # if we are resuming a component, the inputs are already consumed, so we just return the first input
+            if is_resume:
+                consumed_inputs[socket_name] = socket_inputs[0]
+            elif socket_inputs:
```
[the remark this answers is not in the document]
sjrl: Overall I think this is okay for now, but I may want to revisit this in the future. It just feels a bit odd that we need to pass an `is_resume` flag here but I understand why. 

I'd need more time to look into this to see if there is a way we could avoid this.

haystack_experimental/core/pipeline/pipeline.py
```diff
@@ -206,98 +204,115 @@ def run(  # noqa: PLR0915, PLR0912
...
+                        # we use dict instead of to_dict() because it strips away class types of init params
+                        init_params = {
+                            key: value
+                            for key, value in component["instance"].__dict__.items()
+                            if not key.startswith("__")
+                        }
+
+                        if "_template_string" in init_params:
+                            init_params["template"] = init_params["_template_string"]
+                            init_params.pop("_template_string")
```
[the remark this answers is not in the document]
Amnah199: I am assuming in other components we maintain a rule of keeping the names of params similar in class vars. In that case I can update the `PromptBuilder` to mitigate this scenario.

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+from haystack_experimental.dataclasses import GeneratedAnswer
+
+logger = logging.getLogger(__name__)
+
+
+def _validate_breakpoint(pipeline_breakpoint: Tuple[str, Optional[int]], graph: Any) -> Tuple[str, int]:
+    """
+    Validates the pipeline_breakpoint passed to the pipeline.
+
+    Make sure they are all valid components registered in the pipeline,
```
sjrl, line 25:
```suggestion
    Makes sure the breakpoint contains a valid components registered in the pipeline.
```

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+    If the visit is not given, it is assumed to be 0, it will break on the first visit.
+
+    :param pipeline_breakpoint: Tuple of component name and visit count at which the pipeline should stop.
+    :returns:
+        Tuple of valid pipeline_breakpoint.
```
sjrl, line 30:
```suggestion
        Tuple of component name and visit count representing the `pipeline_breakpoint`
```

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+        else None
+    )
+    return valid_breakpoint
+
+
+def _validate_pipeline_state(resume_state: Dict[str, Any], graph: Any) -> None:
+    """
+    Validates that the resume_state contains valid configuration for the current pipeline.
+
+    Raises a PipelineRuntimeError if any component is missing or if the state structure is invalid.
```
sjrl, line 47:
```suggestion
    Raises a PipelineInvalidResumeStateError if any component is missing or if the state structure is invalid.
```

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+    if components_in_state != components_in_order:
+        raise ValueError(
+            f"Inconsistent state: components in pipeline_state['component_visits'] {components_in_state} "
+            f"do not match components in ordered_component_names {components_in_order}"
+        )
+
+    logger.info("Passed resume state validated successfully.")
+
+
+def load_state(file_path: Union[str, Path]) -> Dict[str, Any]:
```
sjrl, line 120: Especially since we have `State` in `Agent` now I think we should think about a more specific name than `state` in the function name. I see in the docs we call it "pipeline state", but I believe we are really loading a dict with keys "input_data", "pipeline_breakpoint", "pipeline_state" so it feels a bit odd to call the whole thing `pipeline_state` if there is a key called `pipeline_state` inside. 

Do you have an idea on some better namings we could use?

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+    :param component_visits: The visit count of the component that triggered the breakpoint.
+    :param callback_fun: A function to call with the saved state.
+    :param debug_path: The path to save the state to.
+    :param original_input_data: The original input data.
+    :param ordered_component_names: The ordered component names.
+    :raises:
+        Exception: If the debug_path is not a string or a Path object, or if saving the JSON state fails.
+
+    :returns:
+        The saved state dictionary
```
sjrl, line 173: Could we also list all the keys and their respective values in this docstring like we do in the `:returns:` for our components?

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+        f"Resuming pipeline from component: {resume_state['pipeline_breakpoint']['component']} "
+        f"(visit {resume_state['pipeline_breakpoint']['visits']})"
+    )
+
+
+def _validate_resume_state(state: Dict[str, Any]) -> None:
+    """
+    Validates the loaded pipeline state.
+
+    Ensures that the state contains required keys: "input_data", "pipeline_breakpoint", and "pipeline_state".
```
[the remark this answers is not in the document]
Amnah199: `component_visits` are provided inside the `pipeline_state` and we check that in `validate_pipeline_state`. With the next PR I was gonna look into merging these two methods perhaps, so i would leave this for now

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -0,0 +1,317 @@
...
+
+    logger.info(f"Successfully loaded pipeline state from: {file_path}")
+    return state
+
+
+def _save_state(
+    inputs: Dict[str, Any],
+    component_name: str,
+    component_visits: Dict[str, int],
+    callback_fun: Optional[Callable[..., Any]] = None,
```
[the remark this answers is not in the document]
Amnah199: One reason for removing it could be that we havent tested this at all.

haystack_experimental/core/pipeline/breakpoint.py
Amnah199: Hmm if we opt for this then we consistently use `pipeline_snapshot` instead of state everywhere. Including the `resume_state` param for `run` method

sjrl: I think this could make sense. I believe `pipeline_snapshot` is more informative than `resume_state` which from it's name sounds like it should be a boolean

haystack_experimental/core/pipeline/breakpoint.py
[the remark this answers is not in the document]
Amnah199: Created issue here: https://github.com/deepset-ai/haystack/issues/9510

haystack_experimental/core/pipeline/breakpoint.py
```diff
@@ -7,27 +7,28 @@
...
 logger = logging.getLogger(__name__)
 
 
-def _validate_breakpoint(pipeline_breakpoint: Tuple[str, Optional[int]], graph: Any) -> Tuple[str, int]:
+def _validate_breakpoint(pipeline_breakpoint: Tuple[str, Optional[int]], graph: MultiDiGraph) -> Tuple[str, int]:
     """
     Validates the pipeline_breakpoint passed to the pipeline.
 
-    Make sure they are all valid components registered in the pipeline,
+    Makes sure the breakpoint contains a valid components registered in the pipeline.
```
sjrl, line 26:
```suggestion
    Makes sure the breakpoint contains a valid component registered in the pipeline.
```

sjrl (approved):
One small comment, otherwise looks good!

PR #319, opened 2025-06-06
feat: add `SentenceTransformersDocumentImageEmbedder`

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9265

### Proposed Changes:

- add `SentenceTransformersDocumentImageEmbedder`: a component that loads images from Documents (paths are stored in metadata) and embeds them using a multimodal embedding model
- minor changes to utility functions; I am also renaming them to make them internal

### How did you test it?
CI; several new tests
Usage example below.

### Notes for the reviewer
- I tried to implement what we discussed in the POC
- I'm leaving out the Cohere embedder. I developed it only to check whether the interface could work well.

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/generators/chat/amazon_bedrock.py
+++ b/haystack_experimental/components/generators/chat/amazon_bedrock.py
@@ -142,9 +142,9 @@ else:
         return system_prompts, repaired_bedrock_formatted_messages
 
     # NOTE: monkey patches needed to use the new ChatMessage dataclass and _format_messages function
-    original_utils.ChatMessage = ChatMessage
-    original_chat_generator.ChatMessage = ChatMessage
-    original_chat_generator._format_messages = _format_messages
+    original_utils.ChatMessage = ChatMessage  # type: ignore[misc]
+    original_chat_generator.ChatMessage = ChatMessage  # type: ignore[misc]
+    original_chat_generator._format_messages = _format_messages  # type: ignore[assignment]
 
     @component
     class AmazonBedrockChatGenerator(original_chat_generator.AmazonBedrockChatGenerator):  # type: ignore[no-redef]


anakin87:
### Text-Image retrieval example

```python
import glob

from haystack import Document
from haystack.components.embedders import SentenceTransformersTextEmbedder
from haystack.components.retrievers.in_memory import InMemoryEmbeddingRetriever
from haystack.document_stores.in_memory import InMemoryDocumentStore

from haystack_experimental.components.embedders.sentence_transformers_doc_image_embedder import (
    SentenceTransformersDocumentImageEmbedder,
)

text_embedder = SentenceTransformersTextEmbedder(
    model="sentence-transformers/clip-ViT-B-32",
)
text_embedder.warm_up()

# the two embedders share the same embedding backend
image_embedder = SentenceTransformersDocumentImageEmbedder(
    model="sentence-transformers/clip-ViT-B-32",
    file_path_meta_field="image_path",
    size=(50, 50),
)
image_embedder.warm_up()


docs = []
for i, image_path in enumerate(glob.glob("test/test_files/images/*")):
    docs.append(
        Document(
            content=f"Doc {i}",
            meta={"image_path": image_path},
        )
    )

docs_with_embeddings = image_embedder.run(docs)["documents"]

document_store = InMemoryDocumentStore(embedding_similarity_function="cosine")
document_store.write_documents(docs_with_embeddings)

retriever = InMemoryEmbeddingRetriever(
    document_store=document_store,
)

query = "apple"
query_embedding = text_embedder.run(query)["embedding"]
results = retriever.run(query_embedding)
print(results)
```

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,335 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import mimetypes
```
sjrl, line 5: I wonder if like our converters if we should also make a separate `image_embedders` folder. What do you think?

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,335 @@
...
+                local_files_only=self.local_files_only,
+                model_kwargs=self.model_kwargs,
+                tokenizer_kwargs=self.tokenizer_kwargs,
+                config_kwargs=self.config_kwargs,
+                backend=self.backend,
+            )
+            if self.tokenizer_kwargs and self.tokenizer_kwargs.get("model_max_length"):
+                self._embedding_backend.model.max_seq_length = self.tokenizer_kwargs["model_max_length"]
+
+    def _validate_image_paths(self, documents: List[Document]) -> List[Dict[str, str]]:
```
sjrl, line 197: Do you think it's worth making this a utility function so we could reuse it as a utility in the `DocumentToImageContent` component?

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,335 @@
...
+                images_paths_types.append({"path": resolved_file_path, "type": "image"})
+
+        return images_paths_types
+
+    @staticmethod
+    def _process_pdf_documents(
+        images_to_embed: List,
+        pdf_documents: List[Dict[str, Any]],
+        size: Optional[Tuple[int, int]],
+    ) -> None:
```
sjrl, line 250: I think it would be better practice/easier to read to have this function return `images_to_embed` even we are still modifying it in place. Or better yet just have this function return an `images_to_embed` based on `pdf_documents` and handle the update to the global `images_to_embed` where this function is called.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,335 @@
...
+                page_number = doc.meta.get("page_number")
+                if page_number is None:
+                    raise ValueError(
+                        f"Document with ID '{doc.id}' comes from the PDF file '{resolved_file_path}' but is missing "
+                        f"the 'page_number' key in its metadata. Please ensure that PDF documents you are trying to "
+                        f"convert have this key set."
+                    )
+                images_paths_types.append({"path": resolved_file_path, "type": "pdf", "page_number": page_number})
+            else:
+                images_paths_types.append({"path": resolved_file_path, "type": "image"})
```
sjrl, line 241: I realize it's probably a bit more clunky to work with but I wonder if we should store `mimetype` instead of a new `type`. Then at run time if the `mimetype` is in the supported `IMAGE_MIME_TYPES` then we can directly load using PIL.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,335 @@
...
+
+        images_paths_types = self._validate_image_paths(documents)
+
+        images_to_embed: List = [None] * len(documents)
+        pdf_documents = []
+
+        for doc_idx, image_path_type in enumerate(images_paths_types):
+            if image_path_type["type"] == "image":
+                # Process images directly
+                image: Union["Image", "ImageFile"] = PILImage.open(image_path_type["path"])
```
sjrl, line 303: Part of this [comment](https://github.com/deepset-ai/haystack-experimental/pull/319/files#r2137082791). I think we should keep using `IMAGE_MIME_TYPES` as the way to determine if a file is an "image" or not.

test/components/embedders/test_sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,396 @@
...
+
+        validated_paths = embedder._validate_image_paths(documents)
+        assert len(validated_paths) == len(documents)
+        for path_info in validated_paths:
+            assert str(path_info["path"]) in image_paths
+            assert path_info["type"] in ["image", "pdf"]
+            if path_info["type"] == "pdf":
+                assert path_info["page_number"] == 1
+            else:
+                assert "page_number" not in path
```
sjrl, line 322: I think this is wrong? `path` is a for loop variable so I think you meant `path_info["path"]`?

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
sjrl: > so LMK if you see good reasons to put them in a separate folder.

Mostly for consistency with `image_converters` and I'm thinking long-term when we add support for other modalities like `Audio` and maybe `Video` that then the `embedders` folder could get crowded  and separating things out would have been nice.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,335 @@
...
+            images_to_embed,  # type: ignore[arg-type]
+            batch_size=self.batch_size,
+            show_progress_bar=self.progress_bar,
+            precision=self.precision,
+            **(self.encode_kwargs if self.encode_kwargs else {}),
+        )
+
+        docs_with_embeddings = []
+        for doc, emb in zip(documents, embeddings):
+            copied_doc = copy(doc)
```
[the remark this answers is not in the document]
sjrl: Ahh okay I see. If you run the embedder in `Pipeline` then the documents are not modified in place because we always deepcopy inputs and outputs between components. 

I think this is a relevant point to discuss, whether components should guarantee not modifying things in place or if this should only be reliably done at the Pipeline level. Maybe worth opening an issue so we can talk about it as a team? We already have this issue open https://github.com/deepset-ai/haystack/issues/9349 which is tangentially related which is that all the deepcopying we do definitely causes latency issues.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
anakin87: At this point, we determined if the file is an image or a PDF using mimetypes (or we raised an error).
Since we already know this information, I find it clearer to temporarily store this type information and take actions based on that. But I might miss some relevant aspects...

sjrl: Yeah fair enough. At first I thought we were also exposing this information in the component's output but I can see this is completely internal. Then totally fine to leave as is.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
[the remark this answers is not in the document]
anakin87: I opened an issue: https://github.com/deepset-ai/haystack/issues/9505

For this experimental component, do you think that we should keep the default (potentially confusing) behavior?

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
anakin87: It makes sense, but I would do this in a subsequent PR to keep the current one relatively focused.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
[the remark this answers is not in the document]
sjrl: No let's leave the `copy` in there.

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
sjrl: sounds good!

haystack_experimental/components/embedders/sentence_transformers_doc_image_embedder.py
anakin87: done in https://github.com/deepset-ai/haystack-experimental/pull/319/commits/a9068d9c44fd24a2104672b3066733c4a1564220

haystack_experimental/components/embedders/image/sentence_transformers_doc_image_embedder.py
```diff
@@ -0,0 +1,361 @@
...
+)
+from haystack_experimental.dataclasses.image_content import IMAGE_MIME_TYPES
+
+with LazyImport("Run 'pip install pillow'") as pillow_import:
+    from PIL import Image as PILImage
+    from PIL.Image import Image
+    from PIL.ImageFile import ImageFile
+
+
+class ImageSourceInfo(TypedDict):
```
sjrl, line 35: Is there a way to make this private?

anakin87: I thought about this point.
If we don't export this to `__init__.py` and we only use it in internal methods I would not expect users to try to use it.
But if we want to be more explicit, we can rename it to `_ImageSourceInfo`

sjrl: Hmm I think we if we'd really like it to be private we should make it that way. Another way I've seen libraries do this (e.g. datadog) is to make the import path include a `_`. So we could do something like `../image/_types.py` or some other file name to indicate things inside that file are private.

anakin87: For now, I just renamed the `TypedDict`s in https://github.com/deepset-ai/haystack-experimental/pull/319/commits/49dc95aa22a6d0ecf7103d0f3f34b13227a8c5b2.

If we end up reusing them in other modules (https://github.com/deepset-ai/haystack-experimental/pull/319#discussion_r2137461451), I will also move them.

anakin87:
> I also wanted to generally ask about how these image embedders handle different image sizes? Does sentence transformers do some automatic resizing or do most image embedders nowadays accept any resolution size?

Thanks for the question, which threw me down quite a rabbit hole.

Sentence Transformers is based on Transformers, which does resizing and rescaling by default.
Sentence Transformers does not allow customizing this behavior.

For CLIP:
- https://github.com/UKPLab/sentence-transformers/blob/deb8e8d3b74d99645686a8760c7ca459b38a5582/sentence_transformers/models/CLIPModel.py#L25C39-L25C52
- https://github.com/huggingface/transformers/blob/71f73859423f12888dcd460ae4ff1fcbb73d5625/src/transformers/models/clip/image_processing_clip.py#L59

For jina-clip-v2:
- https://huggingface.co/jinaai/jina-clip-v2/blob/main/custom_st.py#L123
- https://huggingface.co/jinaai/jina-clip-v2/blob/main/preprocessor_config.json

Depending on how a specific model is implemented, these details can vary.

What emerged is that **resizing/rescaling is already happening** and can be costly (https://github.com/UKPLab/sentence-transformers/issues/1200), so I would be inclined not to expose the `size` parameter in this component, to avoid duplicate processing. For other embedders with specific requirements, we could consider exposing this parameter to simplify the user experience.

WDYT?

anakin87:
Test failures are unrelated. I'm fixing them in #322

PR #330, opened 2025-06-18
feat: breakpoints structure

davidsbatista:
### Related Issues

- Fixes [9533](https://github.com/deepset-ai/haystack/issues/9533)

### Proposed Changes:

- adding data classes to represent different types of breaking points

### Setting Breakpoints in a Pipeline with an Agent 
```python

from haystack_experimental.core.pipeline import Pipeline

chat_breakpoint = Breakpoint(component_name="chat_generator ", visit_count=0)
tool_breakpoint_1 = ToolBreakpoint(component_name="tool_invoker", visit_count=0, tool_name="web_search_tool")
tool_breakpoint_2 = ToolBreakpoint(component_name="tool_invoker", visit_count=0, tool_name="parse_web_results_tool")
agent_breakpoints = AgentBreakpoint(breakpoints={chat_breakpoint,tool_breakpoint_1, tool_breakpoint_2})

breakpoint = Breakpoint(component_name="query_embedder", visit_count=0)
breakpoint2 = Breakpoint(component_name="retriever", visit_count=3)
breakpoint3 = Breakpoint(component_name="chat_generator", visit_count=0)

pipeline_breakpoints = {breakpoint,breakpoint2,breakpoint3, agent_breakpoints}


pipeline = Pipeline()
data = {"query": "What is the capital of France?"}
pipeline.run(data, breakpoints=pipeline_breakpoints)
```

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/dataclasses/breakpoints.py
+++ b/haystack_experimental/dataclasses/breakpoints.py
@@ -47,7 +47,7 @@ class ToolBreakpoint(Breakpoint):
     def __eq__(self, other):
         if not isinstance(other, ToolBreakpoint):
             return False
-        return super().__eq__(other)
+        return super().__eq__(other) and self.tool_name == other.tool_name
 
     def __str__(self):
         if self.tool_name:
@@ -85,6 +85,9 @@ class AgentBreakpoint:
 
         if not breakpoints:
             raise ValueError("Breakpoints must be provided.")
+
+        self.tool_breakpoints = set()
+        self.generator_breakpoints = set()
 
         for break_point in breakpoints:
             if isinstance(break_point, Breakpoint):


haystack_experimental/dataclasses/breakpoints.py
```diff
@@ -0,0 +1,93 @@
...
+@dataclass
+class ToolBreakpoint(Breakpoint):
+    """
+    A dataclass to hold a breakpoint that can be used to debug a Tool.
+
+    If tool_name is None, it means that the breakpoint is for every tool in the component.
+    Otherwise, it means that the breakpoint is for the tool with the given name.
+    """
+
+    tool_name: Optional[str] = None
```
davidsbatista, line 42: probably here we want a `Optional[List[str]]` to set breakpoints at multiple tools

haystack_experimental/dataclasses/breakpoints.py
```diff
@@ -0,0 +1,94 @@
...
+
+    It holds a set of breakpoints for the components in the Agent.
+    """
+
+    breakpoints: Set[Union[Breakpoint, ToolBreakpoint]]
+
+    def __init__(self, breakpoints: Set[Union[Breakpoint, ToolBreakpoint]]):
+        if breakpoints is None:
+            breakpoints = set()
+        self.breakpoints = breakpoints
```
sjrl, line 82: I think it would be nice if we could also verify that the component_name is either `chat_generator` or `tool_invoker` in the respective breakpoints and throw an error if they don't match. This way we can give an early error to the user to fix their breakpoints. What do you think?

haystack_experimental/dataclasses/breakpoints.py
```diff
@@ -0,0 +1,94 @@
...
+
+@dataclass
+class AgentBreakpoint:
+    """
+    A dataclass to hold a breakpoint that can be used to debug an Agent.
+
+    It holds a set of breakpoints for the components in the Agent.
+    """
+
+    breakpoints: Set[Union[Breakpoint, ToolBreakpoint]]
```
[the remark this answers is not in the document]
sjrl: > I don't think there's any hard limit on the number of breakpoints, as many as tools x number of possible visits.

Ahh right good point, hadn't considered how `visit_count` could lead to wanting more breakpoints for the same component

haystack_experimental/dataclasses/breakpoints.py
davidsbatista: agree

haystack_experimental/dataclasses/breakpoints.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if any(isinstance(bp, ToolBreakpoint) and bp.component_name != "tool_invoker" for bp in breakpoints):
+            raise ValueError("All ToolBreakpoints must have component_name 'tool_invoker'.")
+
+        if not breakpoints:
+            raise ValueError("Breakpoints must be provided.")
+
+        for break_point in breakpoints:
+            if isinstance(break_point, Breakpoint):
+                self.generator_breakpoints.add(break_point)
```
Amnah199, line 91: This will throw an error as we haven't initialized these class vars:
```
self.generator_breakpoints = set()
self.tool_breakpoints = set()
```

davidsbatista: forgot those, thanks

haystack_experimental/dataclasses/breakpoints.py
```diff
@@ -0,0 +1,125 @@
...
+
+    tool_name: Optional[str] = None
+
+    def __hash__(self):
+        return hash((self.component_name, self.visit_count, self.tool_name))
+
+    def __eq__(self, other):
+        if not isinstance(other, ToolBreakpoint):
+            return False
+        return super().__eq__(other)
```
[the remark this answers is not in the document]
davidsbatista: good catch 👍🏽

haystack_experimental/dataclasses/breakpoints.py
```diff
@@ -0,0 +1,125 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import dataclass
+from typing import Optional, Set, Union
+
+
+@dataclass
```
Amnah199, line 9: Maybe a good usage of `@dataclass(frozen=True)` that will create `hash` and `eq` automatically?
I dont expect that we will change or update the `Breakpoint` instances

Amnah199:
> pipeline_breakpoints = {breakpoint,breakpoint2,breakpoint3, agent_breakpoints}

@davidsbatista Just out of curiosity do we have a plan for implementing multiple breakpoints. This would be the major refactor to what we currently have.

davidsbatista:
> @davidsbatista Just out of curiosity do we have a plan for implementing multiple breakpoints. This would be the major refactor to what we currently have.

no - but I don't think it stops us from moving to this dataclass even if they supports it - it will break on the first breakpoint and dump everything

haystack_experimental/dataclasses/breakpoints.py
davidsbatista: only thing is that than any class that inherits form Breakpoint also needs to have `frozen=True`

Amnah199:
Your dataclasses are for sure good! I was just wondering if we agreed on support multiple breakpoints already because that will require refactoring 🤔

haystack_experimental/dataclasses/breakpoints.py
davidsbatista: I would leave it as is

Amnah199 (approved):

PR #334, opened 2025-06-25
feat: add `DocumentLengthRouter` component

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9512

### Proposed Changes:
- add a very simple component to route `Documents` based on the length of the `content` field

### How did you test it?
CI, new tests

### Checklist

- I documented my code

### DIFF
--- a/haystack_experimental/components/routers/document_length_router.py
+++ b/haystack_experimental/components/routers/document_length_router.py
@@ -48,9 +48,10 @@ class DocumentLengthRouter:
         Initialize the DocumentLengthRouter component.
 
         :param threshold:
-            The threshold for the length of the document `content` field. Documents where `content` is None or whose
-            length is less than or equal to the threshold will be routed to the `short_documents` output.
-            Otherwise, they will be routed to the `long_documents` output.
+            The threshold for the number of characters in the document `content` field. Documents where `content` is
+            None or whose character count is less than or equal to the threshold will be routed to the `short_documents`
+            output. Otherwise, they will be routed to the `long_documents` output.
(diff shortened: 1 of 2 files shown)


haystack_experimental/components/routers/document_length_router.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(
+        self,
+        *,
+        threshold: int = 10,
+    ) -> None:
+        """
+        Initialize the DocumentLengthRouter component.
+
+        :param threshold:
```
sjrl, line 50: Let's specify what unit of length threshold applies to, which appears to be number of characters right?

haystack_experimental/components/routers/document_length_router.py
```diff
@@ -0,0 +1,82 @@
...
+    # {
+    #     "short_documents": [Document(content="Short", ...)],
+    #     "long_documents": [Document(content="Long document ...", ...)],
+    # }
+    """
+
+    def __init__(
+        self,
+        *,
+        threshold: int = 10,
```
sjrl, line 45: I was going to say maybe we could make this `Optional[int]` so setting this to None means we only check the None-ness of a Document's content, but this can basically be achieved with setting threshold to 0 as well right?

sjrl (approved):
Just a few comments, otherwise looks good!

haystack_experimental/components/routers/document_length_router.py
sjrl: I guess we could also leave as is and explain to users you need to set threshold to -1 to only get None docs. I think I'd be fine with either approach.

PR #344, opened 2025-07-07
feat: Add EmbeddingBasedDocumentSplitter component

julian-risch:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/152

### Proposed Changes:

- Add new EmbeddingBasedDocumentSplitter component that is initialized with a TextEmbedder and uses SentenceSplitter internally as described in the issue
- Added unit tests and an integration test

### Notes for the reviewer

We could leave out the SentenceSplitter to simplify the component. I believe it's better to have it.

There is a know limitation that we could work on in a next iteration of this new component. Currently, we always create splits after `sentences_per_group` sentences. For example with `sentences_per_group=3` we split after 3, 6, 9, ... sentences. Even if there is a shift in the topic of the text after sentence 4, we'll split at 3, 6, 9. In a next iteration of this new component, we could use a sliding window approach to split after sentence 4 in that case.

Test for macos are failing. I am still investigating why. Might have to do with nltk under macos?
```
test/components/preprocessors/test_embedding_based_document_splitter.py::TestEmbeddingBasedDocumentSplitter::test_split_document_with_multiple_topics 
-------------------------------- live log call ---------------------------------
WARNING  haystack.components.preprocessors.sentence_tokenizer:sentence_tokenizer.py:234 No abbreviations file found for en. Using default abbreviations.
Fatal Python error: Segmentation fault
```

### Checklist

- I documented my code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 
 dependencies = [
-  "haystack-ai>=2.15.0",
+  "haystack-ai",
 ]
 
 [project.urls]


haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
```diff
@@ -0,0 +1,376 @@
...
+
+    def _cosine_distance(self, embedding1: List[float], embedding2: List[float]) -> float:
+        """
+        Calculate cosine distance between two embeddings.
+        """
+        vec1 = np.array(embedding1)
+        vec2 = np.array(embedding2)
+
+        norm1 = np.linalg.norm(vec1)
+        norm2 = np.linalg.norm(vec2)
```
sjrl, line 242: Does this mean we have an implicit numpy dependency? It normally gets installed with things like transformers, torch or sentence transformers but it's not part of the base haystack deps. 

And since some of our embedders use APIs it's not guaranteed that they will come with a numpy dep so we may want to explicitly include it or see if we could avoid using numpy.

haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
```diff
@@ -0,0 +1,377 @@
...
+
+        return groups
+
+    def _calculate_embeddings(self, sentence_groups: List[str]) -> List[List[float]]:
+        """
+        Calculate embeddings for each sentence group.
+        """
+        embeddings = []
+        for group in sentence_groups:
+            result = self.text_embedder.run(group)
```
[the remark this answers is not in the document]
julian-risch: Thanks, that makes sense! I updated the code to use DocumentEmbedder instead of TextEmbedder.

haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
julian-risch: We do have numpy as a dependency https://github.com/deepset-ai/haystack/blob/2693f39e4430cd83eb6f2860a399b27f6070645a/pyproject.toml#L57

also using it here
https://github.com/deepset-ai/haystack/blob/2693f39e4430cd83eb6f2860a399b27f6070645a/haystack/utils/misc.py#L7

sjrl: Ahh you're right sorry about that. I initially saw it in test deps so I thought that meant it wasn't in main deps. But the testing is only for ensuring we test on version 2 or higher

haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
```diff
@@ -0,0 +1,404 @@
...
+    splitter.warm_up()
+    result = splitter.run(documents=[doc])
+    ```
+    """
+
+    def __init__(
+        self,
+        *,
+        document_embedder: DocumentEmbedder,
+        sentences_per_group: int = 1,
```
julian-risch, line 63: We probably want to increase the default to 3.

julian-risch:
> There is an issue with extra whitespaces being added to sentences when they are grouped. Input text in document: "The weather today is beautiful. The sun is shining brightly. The temperature is perfect for a walk. There are no clouds and no rain." Output text in document: 'The weather today is beautiful. The sun is shining brightly. The temperature is perfect for a walk. '

fixed now and there is a test for it

haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
```diff
@@ -0,0 +1,402 @@
...
+
+        if self.max_length <= self.min_length:
+            raise ValueError("max_length must be greater than min_length.")
+
+    def warm_up(self):
+        """
+        Warm up the component by initializing the sentence splitter.
+        """
+        nltk_imports.check()
+        self.sentence_splitter = SentenceSplitter(
```
sjrl, line 118: Not entirely sure if the `nltk_imports.check()` is needed here since it will be called by `SentenceSplitter` when it is initialized.

haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
```diff
@@ -0,0 +1,402 @@
...
+        return {"documents": split_docs}
+
+    def _split_document(self, doc: Document) -> List[Document]:
+        """
+        Split a single document based on embedding similarity.
+        """
+        if self.sentence_splitter is None:
+            raise RuntimeError("Sentence splitter is not initialized. Call warm_up() first.")
+        if doc.content is None:
+            raise ValueError("Document content cannot be None.")
```
sjrl, line 177: Not entirely sure if these checks are needed since these are checked for already in the `run` method before this is called. Especially since this is a private method I'd be okay with removing these checks, but also fine to leave.

haystack_experimental/components/preprocessors/embedding_based_document_splitter.py
```diff
@@ -0,0 +1,402 @@
...
+        final_splits = self._post_process_splits(splits)
+
+        return self._create_documents_from_splits(final_splits, doc)
+
+    def _group_sentences(self, sentences: List[str]) -> List[str]:
+        """
+        Group sentences into groups of sentences_per_group.
+        """
+        if self.sentences_per_group == 1:
+            return [s.strip() for s in sentences]
```
sjrl, line 200: Is there a reason we need to run `strip` here?

julian-risch: removed

PR #332, opened 2025-06-24
Refactor serialization of breakpoints

Amnah199:
### Related Issues

- fixes #9463

### Proposed Changes:

- Updated serialization logic to use new utility functions.

- Removed redundant serialization tests meant for testing utility functions.

- Renamed and refactored `validate_pipeline_state` to `_validate_components_against_pipeline` for clarity. This function remains separate from `validate_resume_state` because it performs validations that require access to the pipeline graph—ensuring no invalid components are included during a pipeline resume. Without this method, the resuming throws an error with the invalid node, but doesnt provide any explanation.

Note:

- Renaming of resume_state to pipeline_snapshot was not addressed in this PR to prevent major conflicts with David's PRs. We can do that once Agent breakpoints is merged.

PR #351, opened 2025-08-07
chore: unpin Haystack

anakin87:
### Related Issues

When testing core integrations, I noticed an unexpected dependency structure:
- Haystack depends on Haystack experimental
- Haystack experimental depends on Haystack >=2.15.0 (pin introduced in https://github.com/deepset-ai/haystack-experimental/pull/332)

This setup reduces the effectiveness of our lower-bounds testing in core integrations:
even when we attempt to test Haystack against older supported versions, the dependency on experimental forces an upgrade to >=2.15.0, overriding those bounds.

This has a few implications:
- It limits flexibility for downstream users and complicates dependencies resolution
- The pin is not necessary, especially now that the breakpoint feature has been merged into Haystack.

### Proposed Changes:
- unpin haystack

### How did you test it?
CI

### Notes for the reviewer
To actually make this change effective, we would need to release a new experimental version.

### Checklist

- I documented my code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 
 dependencies = [
-  "haystack-ai>=2.15.0",
+  "haystack-ai",
 ]
 
 [project.urls]


sjrl (approved):
Good catch!

PR #353, opened 2025-08-11
feat: Add EmbeddingBasedDocumentSplitter component

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/152

### Proposed Changes:

- Add new EmbeddingBasedDocumentSplitter component that is initialized with a TextEmbedder and uses SentenceSplitter internally as described in the issue
- Added unit tests and an integration test

### How did you test it?

- unit tests, integration tests and a manual verification

### Notes for the reviewer

There is a know limitation that we could work on in a next iteration of this new component. Currently, we always create splits after `sentences_per_group` sentences. For example with `sentences_per_group=3` we split after 3, 6, 9, ... sentences. Even if there is a shift in the topic of the text after sentence 4, we'll split at 3, 6, 9. In a next iteration of this new component, we could use a sliding window approach to split after sentence 4 in that case.

I'm investigating the integration test failing on macOS with a segault

### Checklist

- I documented my code


anakin87:
Based on https://github.com/deepset-ai/haystack-experimental/pull/344, I think @sjrl has more context about this feature.
If you agree, I would ask him to do the review.

davidsbatista:
stil investigating why integration tests fail on macos

davidsbatista:
this is the issue: https://productionresultssa18.blob.core.windows.net/actions-results/9e2fef8f-a1b9-49b4-bfac-a74e527a303d/workflow-job-run-6bf91f44-bfae-5d55-8f23-c59a315785ad/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-08-11T16%3A34%3A14Z&sig=waeC8C50K0rUueTedY6IDh0kNkP2L%2FRBh3Lkv2s1AeQ%3D&ske=2025-08-12T01%3A59%3A09Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-08-11T13%3A59%3A09Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-08-11T16%3A24%3A09Z&sv=2025-05-05

test/components/preprocessors/test_embedding_based_document_splitter.py
```diff
@@ -0,0 +1,487 @@
...
+
+        # There should be more than one split
+        assert len(split_docs) > 1
+        # Each split should be non-empty and respect min_length
+        for split_doc in split_docs:
+            assert split_doc.content.strip() != ""
+            assert len(split_doc.content) >= 30
+        # The splits should cover the original text
+        combined = " ".join([d.content for d in split_docs]).replace(" ", "")
+        original = text.replace(" ", "")
```
sjrl, line 341: I think it'd better to rewrite the test to this. Otherwise it feels a bit like cheating since we remove all white space in each text which is a bit odd. I tested locally and this version still passes for me
```suggestion
        combined = "".join([d.content for d in split_docs])
        original = text
```

test/components/preprocessors/test_embedding_based_document_splitter.py
```diff
@@ -0,0 +1,487 @@
...
+
+        doc = Document(content=text)
+        result = semantic_chunker.run(documents=[doc])
+        split_docs = result["documents"]
+
+        assert len(split_docs) == 11
+
+        # Verify that the splits cover the original content
+        combined_content = "".join([d.content for d in split_docs])
+        assert combined_content.replace(" ", "").replace("\n", "") == text.replace(" ", "").replace("\n", "")
```
sjrl, line 482: Same comment here. This one doesn't pass locally however, which suggests to me there is a bug somewhere. 
```suggestion
        assert combined_content == text
```
Could we investigate why this fails when doing this comparison? Looking at `combined_content` and `text` locally I can see that it's an issue related to losing information about white spaces and newlines.

sjrl:
Looking good! A few high-level comments:

- Could we add an `__init__.py` file to the `embedders` folder?
- I'm also getting this warning `WARNING  haystack.components.preprocessors.sentence_tokenizer:sentence_tokenizer.py:234 No abbreviations file found for en. Using default abbreviations.` We have included the abbreviations files in Haystack [here](https://github.com/deepset-ai/haystack/tree/main/haystack/data/abbreviations) so somewhat odd that they can't be found. Do you also see the same warning when running the tests locally?
-  It would also be great if we could add an integration test where the text being split has page breaks in it so we could confirm that the page number is being correctly calculated for a long text with page breaks in it. For example, you could take the text from `test_split_large_splits_actually_splits` and add a few page breaks throughout and then check that we are counting pages properly.

davidsbatista:
> The problem is that the Haystack SentenceSplitter is using a relative path from its own location to find abbreviation files:
> 
> ```python
> abbreviations_file = Path(__file__).parent.parent.parent / f"data/abbreviations/{lang}.txt"
> ```
> 
> PosixPath('/Users/dsbatista/haystack-experimental/.venv/lib/python3.11/site-packages/haystack/data/abbreviations/en.txt')

I've open an issue for this https://github.com/deepset-ai/haystack/issues/9708

davidsbatista:
> @davidsbatista just realized I think we are missing a docs config file in `docs/pydoc/config` for `preprocessors_api.yaml`

it's there now

README.md
```diff
@@ -41,19 +41,21 @@ that includes it. Once it reaches the end of its lifespan, the experiment will b
...
+| [`QueryExpander`][5]                  | Query Expansion Component          | October 2025      | None         | None | [Discuss][6] |
+| [`EmbeddingBasedDocumentSplitter`][4] | EmbeddingBasedDocumentSplitter     | August 2025       | None         | None | [Discuss][7] |
 
 [1]: https://github.com/deepset-ai/haystack-experimental/blob/main/haystack_experimental/chat_message_stores/in_memory.py
 [2]: https://github.com/deepset-ai/haystack-experimental/blob/main/haystack_experimental/components/retrievers/chat_message_retriever.py
 [3]: https://github.com/deepset-ai/haystack-experimental/blob/main/haystack_experimental/components/writers/chat_message_writer.py
 [4]: https://github.com/deepset-ai/haystack-experimental/discussions/75
 [5]: https://github.com/deepset-ai/haystack-experimental/blob/main/haystack_experimental/components/query/query_expander.py
 [6]: https://github.com/deepset-ai/haystack-experimental/discussions/346
+[7]: https://github.com/deepset-ai/haystack-experimental/discussions/<placeholder>
```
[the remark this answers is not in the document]
davidsbatista: yes, I did not created it because then it would be immediately public, I will create one now and update the link and merge it after

PR #358, opened 2025-09-01
feat: adding new components `MultiQueryKeywordRetriever` and `MultiQueryEmbeddingRetriever`

davidsbatista:
### Related Issues

- fixes #issue-number

### Checklist

- I documented my code


haystack_experimental/components/retrievers/types/protocol.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import Any, Optional, Protocol, TypeVar
+
+# Ellipsis are needed to define the Protocol but pylint complains. See https://github.com/pylint-dev/pylint/issues/9319.
+# pylint: disable=unnecessary-ellipsis
+
+KeywordRetrieverT = TypeVar("KeywordRetrieverT", bound="KeywordRetriever")
+EmbeddingRetrieverT = TypeVar("EmbeddingRetrieverT", bound="EmbeddingRetriever")
+
+
+class KeywordRetriever(Protocol):
```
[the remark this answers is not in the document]
davidsbatista: I would go with TextRetriever than

haystack_experimental/components/retrievers/multi_query_keyword_retriever.py
```diff
@@ -0,0 +1,157 @@
...
+
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.core.serialization import component_to_dict
+from haystack.utils.deserialization import deserialize_component_inplace
+
+from haystack_experimental.components.retrievers.types import KeywordRetriever
+
+
+@component
+class MultiQueryKeywordRetriever:
```
sjrl, line 16: I think we should also update this name to something like `MultiQueryTextRetriever`. For example, we have a retriever called `MongoDBAtlasFullTextRetriever` from our mongodb integration which follows the protocol but isn't exactly a keyword retriever.

haystack_experimental/components/retrievers/multi_query_embedding_retriever.py
```diff
@@ -0,0 +1,151 @@
...
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.components.embedders.types.protocol import TextEmbedder
+from haystack.core.serialization import component_to_dict
+from haystack.utils.deserialization import deserialize_component_inplace
+
+from haystack_experimental.components.retrievers.types import EmbeddingRetriever
+
+
+@component
+class MultiQueryEmbeddingRetriever:
```
sjrl, line 17: We should add warm_up methods to both of these components, which then calls the warm_up methods of the components they accept in their init methods. For example, here is the one we use in `Agent`
```python
    def warm_up(self) -> None:
        """
        Warm up the Agent.
        """
        if not self._is_warmed_up:
            if hasattr(self.chat_generator, "warm_up"):
                self.chat_generator.warm_up()
            self._is_warmed_up = True
```

davidsbatista: right, I had something similar as just code in the `init()` - but this is the way yes

haystack_experimental/components/retrievers/types/protocol.py
```diff
@@ -0,0 +1,62 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Optional, Protocol, TypeVar
+
+# Ellipsis are needed to define the Protocol but pylint complains. See https://github.com/pylint-dev/pylint/issues/9319.
+# pylint: disable=unnecessary-ellipsis
+
+TextRetrieverT = TypeVar("TextRetrieverT", bound="TextRetriever")
+EmbeddingRetrieverT = TypeVar("EmbeddingRetrieverT", bound="EmbeddingRetriever")
```
sjrl, line 11: I don't think these are needed right?

haystack_experimental/components/retrievers/multi_query_embedding_retriever.py
```diff
@@ -0,0 +1,152 @@
...
+
+    ### Usage example
+
+    ```python
+    from haystack.components.embedders import SentenceTransformersQueryEmbedder
+    from haystack_experimental.components.retrievers import MultiQueryEmbeddingRetriever
+    from haystack_integrations.components.retrievers.opensearch import OpenSearchEmbeddingRetriever
+
+    query_embedder = SentenceTransformersQueryEmbedder(model="sentence-transformers/all-MiniLM-L6-v2")
+    retriever = OpenSearchEmbeddingRetriever(document_store=document_store)
```
sjrl, line 33: This example isn't runnable currently since `document_store` isn't defined. Could we switch it to an `InMemoryDocumentStore` and add some documents to the store to make the example runnable?

haystack_experimental/components/retrievers/multi_query_text_retriever.py
```diff
@@ -0,0 +1,159 @@
...
+
+    ### Usage example
+    ```python
+    from haystack import Document
+    from haystack.components.writers import DocumentWriter
+    from haystack.document_stores.in_memory import InMemoryDocumentStore
+    from haystack.document_stores.types import DuplicatePolicy
+    from haystack.components.retrievers import InMemoryBM25Retriever
+    from haystack_experimental.components.query import QueryExpander
+    from haystack_experimental.components.retrievers.multi_query_retriever import MultiQueryTextRetriever
```
sjrl, line 34:
```suggestion
    from haystack_experimental.components.retrievers.multi_query_text_retriever import MultiQueryTextRetriever
```

haystack_experimental/components/retrievers/multi_query_embedding_retriever.py
```diff
@@ -0,0 +1,152 @@
...
+
+        :param retriever: The embedding-based retriever to use for document retrieval.
+        :param query_embedder: The query embedder to convert text queries to embeddings.
+        :param max_workers: Maximum number of worker threads for parallel processing.
+        """
+        self.retriever = retriever
+        self.query_embedder = query_embedder
+        self.max_workers = max_workers
+        self._is_warmed_up = False
+        self.warm_up()
```
sjrl, line 66: Same here the `warm_up` method shouldn't be called in the init method

sjrl (approved):
Looks good! I see a few other small optimizations like comparing document IDs rather than contents to de-duplicate but we can do that later.

davidsbatista:
> Looks good! I see a few other small optimizations like comparing document IDs rather than contents to de-duplicate but we can do that later.

good point!

PR #391, opened 2025-10-28
Add Mem0 integration - support for Mem0 platform

Amnah199:
### Related Issues

- partially fixes -  [#219](https://github.com/deepset-ai/haystack-private/issues/219)

### Proposed Changes:

- This PR integrates mem0 as a memory store in Haystack, supporting basic add, retrieve, and delete operations.
- It updates the memory_agents.Agent class to enable retrieving and persisting long-term conversational context.

### Key Changes

- Added memory support to memory_agents.Agent.
- Before each execution, relevant memories are fetched from `Mem0MemoryStore` using the latest user message as the query.
- Retrieved memories are injected into the conversation as [MEMORY #n] ChatMessage objects.
- When memories exist, the system prompt is updated to tell the LLM that [MEMORY] messages represent long-term context.
- After each run, new user and assistant messages are saved to the memory store.
- Memory is optional; agents behave normally when memory_store=None.

### How did you test it?

- Unit and Integration Tests
- Tested locally with examples

### Notes for the reviewer

Support for `Memory.from_config` will be revised in a separate PR as it has a slightly different implementation than mem0 `MemoryClient`. 
https://docs.mem0.ai/platform/platform-vs-oss

### Checklist

- I documented my code

### DIFF
--- a/pydoc/memory_store_api.yml
+++ b/pydoc/memory_store_api.yml
@@ -1,0 +1,28 @@
+loaders:
+- ignore_when_discovered:
+  - __init__
+  modules:
+  - haystack_experimental.memory_stores.mem0.memory_store
+  search_path:
+  - ../
+  type: haystack_pydoc_tools.loaders.CustomPythonLoader
+processors:
+- do_not_filter_modules: false
+  documented_only: true
+  expression: null
+  skip_empty_modules: true
+  type: filter
+- type: smart
+- type: crossref
+renderer:
+  description: Storage for the memories using Mem0 as the backend.
+  id: experimental-mem0-memory-store-api
+  markdown:
+    add_member_class_prefix: false
+    add_method_class_prefix: true
+    classdef_code_block: false
+    descriptive_class_title: false
+    descriptive_module_title: true
+    filename: experimental_mem0_memory_store_api.md
+  title: Mem0 Memory Store
+  type: haystack_pydoc_tools.renderers.DocusaurusRenderer


haystack_experimental/components/memory_agents/agent.py
```diff
@@ -0,0 +1,467 @@
...
+                "context and use them to guide the response if relevant."
+            )
+            updated_system_prompt = f"{system_prompt}{memory_instruction}"
+
+            memory_messages = [
+                ChatMessage.from_system(
+                    text=f"[MEMORY #{idx + 1}] {memory.text}",
+                    meta=memory.meta,
+                )
+                for idx, memory in enumerate(retrieved_memory)
```
sjrl: Did you experiment with making these user messages vs system messages? Also what's the reason for making multiple system messages vs one system message with all the memories in it?

pyproject.toml
```diff
@@ -27,8 +27,9 @@ classifiers = [
 ]
 
 dependencies = [
-  "haystack-ai",
+  "haystack-ai==2.20.0",
```
sjrl: Let's make sure not to pin the haystack version. This causes issues when installing things like haystack-core-integrations. Basically pinning haystack-ai here causes other packages that instal haystack to always install this version. @anakin87 knows more about this.

haystack_experimental/memory_stores/mem0/src/mem0/memory_store.py
```diff
@@ -0,0 +1,227 @@
...
+
+from typing import Any, Optional
+
+from haystack import default_from_dict, default_to_dict
+from haystack.dataclasses.chat_message import ChatMessage
+from haystack.lazy_imports import LazyImport
+from haystack.utils import Secret, deserialize_secrets_inplace
+
+with LazyImport(message="Run 'pip install mem0ai'") as mem0_import:
+    from mem0 import Memory, MemoryClient
```
sjrl: We are adding `mem0` as a dependency to the project so no need to put it under a `LazyImport`. Long term I imagine we will make this a core-integration and the `Agent` will only use the to-be-made protocol which we can put into haystack main.

haystack_experimental/components/memory_agents/agent.py
sjrl: Okay interesting. Could you also try with other models like Llama (via OpenRouter or Ollama) and deepseek. I don't think all LLMs do well with multiple system messages being sent. I believe some only expect there to be one system message.

Amnah199: Regardless, I think I can combine this in a single system message. That was something I didnt think of :) 
Will update you after testing some more models.

haystack_experimental/memory_stores/mem0/src/mem0/memory_store.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    A memory store implementation using Mem0 as the backend.
+    """
+
+    def __init__(
+        self,
+        *,
+        user_id: Optional[str] = None,
+        run_id: Optional[str] = None,
+        agent_id: Optional[str] = None,
```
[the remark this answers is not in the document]
Amnah199: I am not sure I understood that completely. 
I think one Agent run should retrieve and update memories for one scope defined by user while initializing memory store instance. Or can you think of cases where Agent would retrieve memories from different scopes (user_ids, agent_ids) in a single conversation? 

Also you mean ids should be supplied at runtime of Agent as `kwargs` or in the methods like `add_memories`, `search_memories`?

haystack_experimental/components/memory_agents/agent.py
```diff
@@ -0,0 +1,467 @@
...
+            span.set_tag("haystack.agent.steps_taken", exe_context.counter)
+
+        # Add the new conversation as memories to the memory store
+        new_memories = [
+            message
+            for message in result["replies"]
+            if message.role.value == "user" or message.role.value == "assistant"
+        ]
+        if self.memory_store:
+            self.memory_store.add_memories(new_memories)
```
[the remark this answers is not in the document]
Amnah199: Updated this to use `result["messages"]`.

haystack_experimental/memory_stores/mem0/src/mem0/memory_store.py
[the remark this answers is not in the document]
Amnah199: Updated!

haystack_experimental/memory_stores/mem0/src/mem0/memory_store.py
```diff
@@ -0,0 +1,227 @@
...
+                        query=search_query,
+                        limit=search_top_k,
+                        filters=mem0_filters,
+                    )
+            if _include_memory_metadata:
+                for memory in memories["results"]:
+                    meta = memory.copy()
+                    meta.pop("memory")
+                    messages = [
+                        ChatMessage.from_user(text=memory["memory"], meta=meta) for memory in memories["results"]
```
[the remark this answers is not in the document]
Amnah199: I updated this to be more clear. `include_memory_metadata` means that we also include the metadata coming from mem0 in a separate key `retrieved_memory_metadata`. This can be helpful if the user wants to access `score`, or `memory_id` etc.

haystack_experimental/memory_stores/mem0/src/mem0/memory_store.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        Delete memory from Mem0.
+
+        :param memory_id: The ID of the memory to delete.
+        """
+        try:
+            self.client.delete(memory_id=memory_id)
+            print(f"Memory {memory_id} deleted successfully")
+        except Exception as e:
+            raise RuntimeError(f"Failed to delete memory {memory_id}: {e}") from e
```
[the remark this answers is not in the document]
sjrl: Okay I understand. Then I'd only say either drop the print statement or convert that into a logger.info statement.

haystack_experimental/memory_stores/mem0/memory_store.py
```diff
@@ -0,0 +1,210 @@
...
+        for message in messages:
+            if not message.text:
+                continue
+            mem0_message = [{"content": message.text, "role": message.role.value}]
+            mem0_metadata = message.meta.copy()
+            # we save the role of the message in the metadata
+            mem0_metadata.update({"role": message.role.value})
+
+            try:
+                result = self.client.add(messages=mem0_message, metadata=mem0_metadata, infer=infer, **ids)
```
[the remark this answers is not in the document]
sjrl: sounds good!

haystack_experimental/components/agents/agent.py
```diff
@@ -135,6 +141,7 @@ def __init__(
         confirmation_strategies: Optional[dict[str, ConfirmationStrategy]] = None,
         tool_invoker_kwargs: Optional[dict[str, Any]] = None,
         chat_message_store: Optional[ChatMessageStore] = None,
+        memory_store: Optional["Mem0MemoryStore"] = None,
```
sjrl: This should use the type defined in `memory_stores.types.protocol`

haystack_experimental/components/agents/agent.py
```diff
@@ -495,7 +552,15 @@ def run(  # type: ignore[override]  # noqa: PLR0915
...
-            result["last_message"] = msgs[-1]
+            result["messages"] = msgs
+            result["last_message"] = msgs[-1] if msgs else None
+
+            # Add the new conversation as memories to the memory store
+            new_memories = [
+                message for message in msgs if message.role.value == "user" or message.role.value == "assistant"
+            ]
+            if self._memory_store:
+                self._memory_store.add_memories(messages=new_memories, **memory_store_kwargs)
```
sjrl: Let's create the new list only if it will be used. Also I think it's simpler to just throw out the system message, since that's what contains the retrieved memories currently right?
```suggestion
            if self._memory_store:
                new_memories = [
                    message for message in msgs if message.role.value != "system"
                ]
                self._memory_store.add_memories(messages=new_memories, **memory_store_kwargs)
```

haystack_experimental/components/agents/agent.py
```diff
@@ -495,7 +552,15 @@ def run(  # type: ignore[override]  # noqa: PLR0915
 
         result = {**exe_context.state.data}
         if msgs := result.get("messages"):
-            result["last_message"] = msgs[-1]
+            result["messages"] = msgs
+            result["last_message"] = msgs[-1] if msgs else None
```
sjrl: Revert back. The `if msgs := result.get("messages"):` already covers this. Using the walrus operator here means that if `msgs` is None then everything under the if statement is skipped
```suggestion
            result["last_message"] = msgs[-1]
```

haystack_experimental/memory_stores/types/protocol.py
```diff
@@ -0,0 +1,99 @@
...
+        ...
+
+    def add_memories(
+        self,
+        *,
+        messages: list[ChatMessage],
+        user_id: Optional[str] = None,
+        run_id: Optional[str] = None,
+        agent_id: Optional[str] = None,
+    ) -> list[str]:
```
sjrl: I think a good middle of the road approach here would be to change this to
```suggestion
    def add_memories(
        self,
        *,
        messages: list[ChatMessage],
        user_id: Optional[str] = None,
        **kwargs: Any,
    ) -> list[str]:
```
so we always require an implementation of MemoryStore at least has a `user_id` but then all other ids that may be specific to the specific backend aren't required for all implementations. 

If you agree let's update all functions in this protocol to work this way

haystack_experimental/memory_stores/mem0/memory_store.py
```diff
@@ -0,0 +1,289 @@
...
+                    messages = [
+                        ChatMessage.from_system(text=memory["memory"], meta=meta) for memory in memories["results"]
+                    ]
+            else:
+                # we only include the metadata stored in the memory in ChatMessage
+                messages = [
+                    ChatMessage.from_system(text=memory["memory"], meta=memory["metadata"])
+                    for memory in memories["results"]
+                ]
+            return messages
```
sjrl: We can simplify this to 
```suggestion
            messages = []
            for memory in memories["results"]:
                meta = memory["metadata"].copy() if memory["metadata"] else {}
                # we also include the mem0 related metadata i.e. memory_id, score, etc.
                if include_memory_metadata:
                    meta["retrieved_memory_metadata"] = memory.copy()
                    meta["retrieved_memory_metadata"].pop("memory")
                messages.append(
                    ChatMessage.from_system(text=memory["memory"], meta=meta)
                )
            return messages
```

haystack_experimental/memory_stores/mem0/memory_store.py
```diff
@@ -0,0 +1,289 @@
...
+    def delete_memory(self, memory_id: str, **kwargs: Any) -> None:
+        """
+        Delete memory from Mem0.
+
+        :param memory_id: The ID of the memory to delete.
+        :param kwargs: Additional keyword arguments to pass to the Mem0 client.delete method.
+        """
+        try:
+            self.client.delete(memory_id=memory_id, **kwargs)
+            logger.info(f"Memory {memory_id} deleted successfully")
```
sjrl:
```suggestion
            logger.info("Memory {memory_id} deleted successfully", memory_id=memory_id)
```

test/memory_stores/test_memory.py
```diff
@@ -0,0 +1,60 @@
...
+    - MEMORY #2: The user's most projects are related to NLP and LLM agents.
+    - MEMORY #3: The user finds it easier to use the Haystack framework to build projects.
+    """)]
+    messages = memory_instruction + user_message + memories
+    return messages
+
+class TestMemory:
+
+    @pytest.mark.integration
+    def test_memory_with_claude(self, messages):
```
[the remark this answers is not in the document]
sjrl: That looks good! I think for bedrock just testing a cohere model would be fine. Testing nova is probably not necessary.

README.md
```diff
@@ -48,6 +48,7 @@ that includes it. Once it reaches the end of its lifespan, the experiment will b
...
 | [`InMemoryChatMessageStore`][1]; [`ChatMessageRetriever`][2]; [`ChatMessageWriter`][3]                                                                                                  | Chat Message Store, Retriever, Writer | February 2025     | None         | <a href="https://colab.research.google.com/github/deepset-ai/haystack-cookbook/blob/main/notebooks/conversational_rag_using_memory.ipynb" target="_parent"><img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/>       | [Discuss][4]  |
+| [`Mem0MemoryStore`][26]                                                                                                                                                                   | MemoryStore                   | February 2025      | mem0ai         | None                                                                                                                                                                                                                                                       | [Discuss][25] |
```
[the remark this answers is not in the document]
sjrl: Also the discussion link is incorrect, it points to the LLM summarizer discussion.

.github/workflows/tests.yml
```diff
@@ -129,5 +133,15 @@ jobs:
...
+      # Do not authenticate on PRs from forks and on PRs created by dependabot
+      - name: AWS authentication
+        id: aws-auth
+        if: github.event_name == 'schedule' || (github.event.pull_request.head.repo.full_name == github.repository && !startsWith(github.event.pull_request.head.ref, 'dependabot/'))
+        uses: aws-actions/configure-aws-credentials@61815dcd50bd041e203e49132bacad1fd04d2708
+        with:
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
+
       - name: Run
```
sjrl: Seems like we also have this in the bedrock integration tests
```suggestion
      - name: Run
        if: success() && steps.aws-auth.outcome == 'success'
```

sjrl:
@anakin87 could we ask for your help on what might be going wrong with the AWS auth setup in tests?

anakin87:
@sjrl try adding
```
permissions:
  id-token: write
  contents: read
```

on the test workflow, on top level
as in https://github.com/deepset-ai/haystack-core-integrations/blob/943ad7f0e7633dd43a0331602ec41c6074fca5c3/.github/workflows/amazon_bedrock.yml#L22-L24

It seems it also worked on this repo in the past

PR #481, opened 2026-06-16
fix: prevent MarkdownHeaderLevelInferrer regex from consuming newlines

JSap0914:
### Related Issues

No existing issue — discovered during code review.

### Proposed Changes

The header-matching regex in `MarkdownHeaderLevelInferrer` used `\s+` between the
`#` characters and the heading text.  `\s` matches *any* whitespace, including
newline characters.  When a header line contained only trailing whitespace (e.g.
`## `), the regex engine could consume the trailing space **and** the following
newline in one `\s+` match, then lazily extend `(.+?)` over the content of the
next header line.  This caused a whitespace-only pseudo-header and the following
real header to be matched as a single span, rewriting:

```
## \n## Section\nContent
```

to the invalid output:

```
# ## Section\nContent
```

(a literal `#` character embedded inside the heading text).

**Fix:** replace `\s+` / `(?:\s*)` with `[ \t]+` / `(?:[ \t]*)` so the
pattern is restricted to horizontal whitespace and cannot span line boundaries.

### How did you test it?

- Two new regression tests added to `test/components/preprocessors/test_markdown_header_level_inferrer.py`.
- All 14 pre-existing tests continue to pass.
- Manual verification:
  ```
  python -m pytest test/components/preprocessors/test_markdown_header_level_inferrer.py -v
  # 16 passed
  ```

### Notes for the reviewer

The change is confined to a single compiled regex literal in `__init__`; no
surrounding logic changes.

### Checklist

- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`
- I documented my code
- I ran pre-commit hooks and fixed any issue

> This PR was fully generated with an AI assistant. I have reviewed the changes and run the relevant tests.


anakin87 (approved):
👍

